PR #32, opened 2014-12-01
Remove circular ProgressBar from ActionBar

vaguelyasian:
This removes the circular ProgressBar from the ActionBar when loading
images/gifs/webms. There already is a custom horizontal ProgressBar
running that's determinate. Having two ProgressBars, one in the
ActionBar, one below the ActionBar, is kind of overkill.

The `boolean showProgressBar` variable remained unused after these changes, so it's removed from code as well.

Any feedback on this is appreciated.


Floens:
This was completely removed with the new controller system, and doesn't apply anymore!
Still thank you very much for your contribution.

PR #33, opened 2014-12-01
Wrap pinned ListView around a parent View

vaguelyasian:
This gives the possibility to swipe on the remaining view (the non-ListView) and tuck the drawer back in. It also helps when you're trying to pull the drawer out, that it does not grab a list-view item and disregard the drawer animation for a swipe dismiss animation.

This might need more eyes/hands-on before getting merged. It works fine on my side after using it for about 20 minutes.

Any feedback on this is appreciated!


Floens:
Like with the other pr, this is not relevant anymore.
Thanks anyway!

PR #42, opened 2015-01-16
Allow downloading of threads from thread view (without opening image view first)

siema:
I'm downloading lots of threads and thought that it's kinda annoying to be forced to open the image everytime. Hopefully it'll help others as well.


Floens:
Thanks!

PR #43, opened 2015-01-18
Material for the material design branch

beschoenen:
I made some fixed for the material branch and added a function to share image links instead of the image itself


Floens:
Why did you close the pr?
Looks all good to me. I changed the toolbar offset to use dp and formatted some code.

beschoenen:
Well i wasn't exactly sure if you wanted all of my commits so i closed it.
Anyway, Thanks for merging them.

Issue #34, opened 2014-12-02
Feature: Undo Pin swipe 


bacevs:
A popup to undo accidental swipe of a pinned topic would be nice

Floens:
Was added.

PR #46, opened 2015-02-01
Snackbar with undo button when you swipe away pins.

Sietsem:
An attempt to create #34. I am using a new library called snackbar to create Snackbars mentioned in the android design specs.

Something that could be improved is the color that is now hardcoded, I wasn't sure where to put it.


Floens:
I merged this manually, with the snackbar provided by the design lib.

PR #47, opened 2015-02-03
Fixed '<>' operator is not allowed for source level below  1.7

LOG-TAG:
Fixed '<>' operator is not allowed for source level below  1.7


Floens:
Clover compiles in 1.7. What's the problem?
Also there are more 1.7 brackets in Clover.

LOG-TAG:
Solved I felt I'm using java 1.7 but it's targeting 1.6 !! Sorry !!

PR #54, opened 2015-03-08
Add thread hiding functionality

allanlw:
This set of commits adds the ability to "hide" threads. The set of hidden threads are stored in a new database table, and the hiding uses roughly the same mechanism as the filtering mechanism. This is a feature that I have wanted in Clover for a while, so I decided to add it myself.

Things added:
  - new DB table for hidden threads
  - new button in board grid / list view for 'hide thread', right next to pin
  - new button for 'reset hidden threads' in context menu, which resets the hidden threads for the current board.

It would be nice to also have a 'show hidden' option, but I wouldn't personally use it, so I didn't add it.

There is also an issue that causes some thumbnails to show up incorrectly in the grid/list views after hiding a thread and then hitting the 'refresh' button. I tried to debug this issue but I am not sure what the cause of it would be. My hunch is that it has something to do with the `CustomNetworkImageView` for the `PostView` not getting reset correctly. This issue needs to be fixed before merging in, and any help would be appreciated.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ChanApplication.java
+++ b/Clover/app/src/main/java/org/floens/chan/ChanApplication.java
@@ -29,6 +29,7 @@ import com.android.volley.toolbox.Volley;
 import org.floens.chan.chan.ChanUrls;
 import org.floens.chan.core.ChanPreferences;
 import org.floens.chan.core.manager.BoardManager;
+import org.floens.chan.core.manager.HideManager;
 import org.floens.chan.core.manager.ReplyManager;
 import org.floens.chan.core.manager.WatchManager;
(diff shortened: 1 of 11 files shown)


Floens:
Very nice, I'll merge it as soon as possible.

allanlw:
Great.

Any ideas about what would cause the image mismatch issue?

Floens:
I'll now manually add this.

Floens:
Added manually in https://github.com/Floens/Clover/commit/7ecc9f183b873438b54f6656e4fd97a88caca3fc
Not exactly the same, I added clearing of hides in the settings, so no clearing per board but globally.

PR #71, opened 2015-05-09
Fix resource leak in ImageSaver.java

nopjmp:
While investigating into problems with Clover crashing on my tablet (looks like a memory leak somewhere in the App Store version) I decided to look at the dev branch and found that there was 4 objects leaking in the ImageSaver. I've also added Square's LeakCanary to my local tree and haven't found any leaks with that yet in the dev branch. I'll open another pull request with that change later if you are interested.


Floens:
Thanks! That's was pretty stupid line of code!
LeakCanary looks pretty cool, I'll also check it out.

PR #73, opened 2015-05-09
Report posts through a dialog

nopjmp:
This allows users to not have to switch back and forth between their browser and Clover just to report a post. Also get rid of ActionBarActivity since this is being deperecated and the new UI doesn't use ActionBarActivity anymore. The padding at the top makes the dialog not look weird. We can always remove the background and make it a floating WebView, but I believe the dialog look is better.


nopjmp:
This type of display could be used for the captcha, but it would require more code.

Floens:
Thanks!

PR #85, opened 2015-07-02
Http proxy settings

KOSBUM:
As you may noticed, this great app has some 1-star raitings on google play with comments stating that it's not working. This is because some internet providers have banned the site in Russia for a single picture, and still keepeng it in ban list despite the fact that this picture isn't exist on the site anymore and the site was removed from official list. Of course, there are several ways to avoid the ban - like using VPN app or system-wide proxy settings. But free VPN & proxy servers are usually slower than your internet connection and slowing all networking apps - browser, email, etc. while growsing 4chan is annoying, so I decided to add proxy settings for the app.


Floens:
Very nice, I'll merge this soon.

Floens:
Thanks! I added caching of the Proxy object and formatted all classes.

PR #90, opened 2015-07-23
Add 'Open in browser' for posts and threads

KingHual:
This adds the 'Open in browser' button back for threads and posts.


Floens:
Thanks! I also added the button to the board view, and added logic to avoid it opening Clover itself.

PR #91, opened 2015-07-26
Add selectable post comments

KingHual:
This makes comments selectable. Apologies if it seems too hacky but I'm not completely comfortable with the code yet.


KingHual:
Noticed you can't click on clickables that way so I'm closing this for now.

Issue #95, opened 2015-08-09
[Proposal] add TinEye to Image Search options


vjandrea:
Hi,
i'd love to see TinEye support in the image search options. I'll try to prepare a PR.

chokladio:
I won't second TinEye but if you're adding more then Saucenao would be a appreciated addition.

https://saucenao.com/

PR #98, opened 2015-08-21
Feature #95 SauceNao

vjandrea:
Added SauceNao image search by @chokladio proposal on issue #95, moved
TinEye on 4th position to mantain the alphabetic order of options.


Floens:
Thanks!

PR #129, opened 2015-11-20
Advanced video loading & FAB auto hiding

beschoenen:
I used [this library](https://github.com/makovkastar/FloatingActionButton) to implement the on scroll listener, I've added the license too.  
  
Though the color of the FAB isn't correct at the moment.

### DIFF
--- a/Clover/app/build.gradle
+++ b/Clover/app/build.gradle
@@ -79,6 +79,7 @@ android {
 }
 
 dependencies {
+    compile 'com.melnykov:floatingactionbutton:1.3.0'
     compile 'com.android.support:support-v13:23.1.0'
     compile 'com.android.support:appcompat-v7:23.1.0'
     compile 'com.android.support:recyclerview-v7:23.1.0'
--- a/Clover/app/src/main/assets/html/licenses.html
+++ b/Clover/app/src/main/assets/html/licenses.html
@@ -194,5 +194,33 @@ The GIFLIB distribution is Copyright (c) 1997  Eric S. Raymond
         </code>
     </pre>
     <br>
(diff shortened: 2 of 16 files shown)


Floens:
Hey thanks for the PR.
I'll probably bind the FAB to the toolbar with hiding, so that they move together.
Why did you change the video loading setting? I can't think of a situation that you do want videos to auto load but not images.

beschoenen:
Video auto loading is still dependent on image auto loading being enabled. But now you chose to still disable video autoloading on your mobile plan.
I have for instance image auto loading on always and video on wifi only.

beschoenen:
He, are you still interested in this PR?, or part of it

Floens:
Yes I am. I'm still busy with the freezing issue. I'll look into this further when that's resolved.

beschoenen:
Okay, thanks for the reply

Floens:
I implemented both of these manually, see my most recent commits.

I've attached the FAB hiding to the toolbar hiding, so that it doesn't require another library.
https://github.com/Floens/Clover/commit/0566c890935adb210705495e4a31f0158fcf8c41

For the video loading on network type, I've renamed the preference to avoid collisions. 
It still depends on the image network type, so the other options are greyed out where appropriate. This way the user still knows that you can't have weird combinations.
For example when you select the image network type to "wifi only", you then can't select the video network type to "all" (wifi+mobile).
The video mode will be set to a lower type automatically when the image mode is changed to something lower.
https://github.com/Floens/Clover/commit/db39d68efa4303d3e51b76b7b9cfdf9843f63fd3

Thank you for the help!

PR #144, opened 2016-01-16
Update the post_details_color so that it can be more easily seen on dark themes

mezpahlan:
On Dark (and inherited) themes, I found the post details and time stamp hard to read for posts that the user has posted (saved posts). The background highlight colour and the post details colour are very similar hence the difficulty in reading. I've borrowed a colour from another theme (Tomorrow) for the `post_details_color` variable - obviously this is very subjective but I feel this strikes a better balance than before. 

### Before
before

### After
after


Floens:
Yeah this is good, thanks!

Issue #155, opened 2016-02-03
If "Save original filename" is disabled and you download a album its still downloads the original name.


WhyANameWhy:
So if "Save original filename" is disabled and you download a album its still downloads the original name. If i activate it nothing changes.

Floens:
Copying my response from the pr #159 
>The reason that album downloads always use the name 4chan provides is because this way the downloader can skip over existing images reliably. If you use the other name and when there are duplicate file names the second one wouldn't be downloaden. I can solve this by doing some md5 checking. But that's the reason why I didn't do this.

There are several options, like inspecting the md5 of the image when it encounters duplicates, or saving what was downloaded in the db.

hardliner5000:
Adding a _1, _2, _3, to the dublicate filename would be possible?

Floens:
Also, the term "original filename" is ambiguous. It could mean the name from the poster of from the server. It's interpreted as the server name now.

>Adding a _1, _2, _3, to the dublicate filename would be possible?
Yes, but that's not the issue here. The single image saver adds _n to the filename if it already existed, but we want the album downloader to skip images it already downloaded. This cannot be done reliably if we use the filename that is user generated. This because someone can post the same "image.jpg" image 5 times and then only the first would be downloaded because it would get seen as already downloaded.
It's now always using the name from 4chan as a workaround, since those are always unique.

onnlinetool:
Downloaden Downloaden Downloaden Downloaden Downloaden Downloaden Downloaden 

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

cm-mc:
Shouldn't that better be left for the user to decide whether to dl chan server filename or the original uploaded filename (using file_n), specially since you can already do that for single file dl?
I was looking for a reader that did this AND allowed to dl all board files at once, and was hoping clover was the one...

CatKasha:
[...]

PR #159, opened 2016-02-08
Album download to use saveOriginalFilename preference

andyklimczak:
Added check for album download to decide which name to use when saving
the file.
Brought up in issue #155

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/chan/ChanParser.java
+++ b/Clover/app/src/main/java/org/floens/chan/chan/ChanParser.java
@@ -133,7 +133,7 @@ public class ChanParser {
             }
         }
 
-        if (!TextUtils.isEmpty(post.name) && !post.name.equals("Anonymous")) {
+        if (!TextUtils.isEmpty(post.name) && (!post.name.equals("Anonymous") || ChanSettings.showAnonymousName.get())) {
             post.nameSpan = new SpannableString(post.name);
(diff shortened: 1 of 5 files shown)


Floens:
The reason that album downloads always use the name 4chan provides is because this way the downloader can skip over existing images reliably. If you use the other name and when there are duplicate file names the second one wouldn't be downloaden. I can solve this by doing some md5 checking. But that's the reason why I didn't do this.

RubarbKid:
You could get around the problem by having a filename map.  Each entry in the map would contain the 4chan assigned name and the associated "original" filename.  I use the quotes because if you had 16 "mycat.jpg" files, you'd be able to rename the redundant files to "mycat+2.jpg" and still keep the essence of the original filename...

RubarbKid:
A bit more overhead than a simple file exists check, but much much faster than an md5 approach...  And not that hard to implement.

RubarbKid:
Sorry, I just realized I left out some of the process...  I didn't mean all 16 files would get "mycat+2.jpg" for a filename.  I meant the 2nd mycat file would get "mycat+2.jpg".  The 3rd would get "mycat+3.jpg" and so on.  The name saved in the map would be the actual saved filename, not the original.   Add a file to the map after a successful download.  The map would only be associated with a single thread and could be kept in the DB (released when the thread is removed), or in a "filename-map.txt" file in the save folder.

andyklimczak:
Good points, I hadn't considered those potential conflicts. I'll close this pr since it has issues. 

I'll submit again if I get inspired with a proper fix.

Issue #148, opened 2016-01-21
Make hiding the default Anonymous name optional


42RK:
some people still prefer to see it

d9h02f:
+1, it feels strange seeing only the dates.

Floens:
Seems you guys have a bit of neophobia ;)
I'll put it on the todo list for an advanced option

PR #161, opened 2016-02-13
Advanced option to show/hide Anonymous username

andyklimczak:
Shows the Anonymous username in post cell when that post is created by anonymous user.
Can be toggled in the advanced options under 'Show Anonymous username'. Defaults to off.

Brought up by #148


Floens:
Seems good, I'll merge this soon.

Floens:
Thanks!

PR #176, opened 2016-04-04
Change repository: mavenCentral -> jcenter

andyklimczak:
Fix 'could not find dependency gradle:2.1.0-alpha5' by switching repository from mavenCentral -> jcenter. 
This fixes the failed build.

I'm not sure if you preemptively updated the version for when mavenCentral finally does get the latest version.


Floens:
Can't hurt, thanks.

Issue #160, opened 2016-02-09
Request: ability to remove highlighting and replies for privacy reasons


1excellent:
I really like clover, it's honestly my favorite app out of all the chan apps available. My only concern is that clover highlights your post and there is no way to clear it and any (you)s will stay until either the thread is deleted or you have to uninstall/reinstall the app. There should be an option to clear out cache/posting history or something like that for privacy reasons because your entire posting history will be visible for as long as the thread exists. I'm aware that clover v2 has history, but it only shows what threads you visited and disabling still keeps the post highlighting.

jocfla:
That's a great idea :D

Floens:
I could make an option to clear the saved reply table. It's a minor feature.

1excellent:
Thank you Floens, it would be wonderful if you could do that.

PR #177, opened 2016-04-04
Add ability to clear saved replies/reply history

andyklimczak:
In a sub-menu located under the 'Clear history' option in the History view.
Requires a thread refresh to take effect

Feature brought up in #160

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseManager.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseManager.java
@@ -282,6 +282,59 @@ public class DatabaseManager {
         return list;
     }
 
+    /**
+     * Clears all stored saved replies
+     */
+    public void clearSavedReplyHistory() {
+        try {
+            TransactionManager.callInTransaction(helper.getConnectionSource(), new Callable<Void>() {
+                @Override
+                public Void call() throws Exception {
+                    List<SavedReply> replyList = getSavedReplyList();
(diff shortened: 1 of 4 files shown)


Floens:
I was just refactoring the databasemanager so give me a few more days before I merge this. The commit itself looks good!

andyklimczak:
I can solve the merge conflicts if you weren't planning on breaking out saved reply into it's own database manager. 

But if you are, I don't think I have the ability to handle that myself.

Floens:
Hey, I merged this manually. I kept it in line with the other db work I've done.

The new system is based on tasks (that's why the new methods all return Callables) that need to be submitted with runTask or runTaskSync. These methods all run the tasks on the same worker thread. You can give runTask a callback if you want to know when the work is done, and runTaskSync waits for the task to finish and then returns. Some work requires waiting for the result and some work can be done without waiting for it to finish. Like clearing the saved replies, that can run without waiting for it to finish.

The saved reply thing is a bit special because the isSaved method is called a lot and we don't want a db call for each call, and we want the call to be thread safe, that's why there is a map to cache all the saved replies and a lock around each statement that modifies these maps.

Move into own class commit https://github.com/Floens/Clover/commit/4a3b9058b888d40b8340a50a5016001213c23b4b
Option to add clearing saved replies https://github.com/Floens/Clover/commit/f97faa136fa24904a562d504c84f74742b4affe6

Thanks!

Issue #182, opened 2016-04-12
Trimming long posts


galenasphaug:
I noticed that while scrolling through the catalog,  posts are trimmed at a set amount of characters,  but not lines. This causes some words to be cut off,  and over all looks a bit strange. It might look a little better if you trimmed the post to about 15 lines.

Floens:
I think the pr is good enough.
The card views have a dynamic height, so setting a max lines is not going to work.
Further, the card views use a faster version of the textview so maxlines does not work out of the box.

PR #184, opened 2016-04-17
Trim long posts

andyklimczak:
Split long thread posts between words instead of in middle of a word
Will fallback to old method if the post is a single long string/spam

(Kind of?) brought up in #182, but I see this in board mode all the time, not catalog mode (don't use catalog mode).


Floens:
Thanks! I didn't know BreakIterator existed, good to know.

PR #192, opened 2016-05-17
Update gradle to 2.1.0

andyklimczak:
Android studio is being annoying, and gets angry about using 2.1.0-alpha5 (for some reason) and sets gradle version to 2.1.0.
Tested with a build and quick smoke test.


Floens:
I usually keep updating this to whatever version of Android Studio I'm using, and set it to the last stable version when I make a release.

Issue #190, opened 2016-05-13
Option to unspoiler thumbnails


riesza:
Is this something that could be done? I'm regularly on a terrible connection and I can't view spoilered  thumbnails without loading the entire image first. With an option like that I could avoid wasting my time with irrelevant spoilered images.

Floens:
Added by the pr of Andy

PR #193, opened 2016-05-20
Add Reveal Image Spoilers in advanced options

andyklimczak:
I think I covered all the cases and got the logic right...
(my manual testing had it working correctly for each scenario)

This option reveals spoiled images in the board/catalog view, thumbnail, album view, and image view. As well as handles the file name and nav title.

Feature requested in #190


Floens:
I'll get to this and the other PR in the weekend.

Floens:
Works great.
It might be neater to change the spoiler boolean all the way up in the parser, but this works too.
Thank you very much!

andyklimczak:
Good point. I understand exactly what you're suggesting. That would make it much cleaner.

I'll do a small refactor for that relatively soon, unless someone else does it !

PR #195, opened 2016-05-23
Add advanced option to reveal spoiler text

andyklimczak:
Gives the spoiler text the appearance that is has already been clicked.

This changes the spoiler text to act similar to 4chanx's spoiler text, where the spoilered text will keep the same background color when revealed, and only the text color changes. This makes sense with the 'Reveal spoiler text' option so that you can still tell the difference between normal text and spoiler text for 'context'.

Also added a 'text_color_reveal_spoiler' since the spoiler text always has the spoiler background, even when revealed. I created a new color instead of using a preexisting color (such as textPrimary) due to some of the text colors on the lighter themes being very similar to spoilerColor (ie: black), making it hard to see when revealed.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/model/PostLinkable.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/model/PostLinkable.java
@@ -22,6 +22,7 @@ import android.text.TextPaint;
 import android.text.style.ClickableSpan;
 import android.view.View;
 
+import org.floens.chan.core.settings.ChanSettings;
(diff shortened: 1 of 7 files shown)


Floens:
Good, I like it more than the old way.
Thank you for your continued support, I greatly appreciate it!

Floens:
(committed with https://github.com/Floens/Clover/commit/71d9b1d3e3af5696c1e0d9b65570363ffba04cdc)

PR #203, opened 2016-06-25
Added optional dual/multi window for Samsung and LG devices

braeden123:
I found this feature super useful for a variety of other apps, and I thought it might also help here. Not tested, but it should work- [reference](https://www.reddit.com/r/Android/comments/3pw4oq/how_to_add_dual_window_support_to_any_app_no_root/cwai1ma)

### DIFF
--- a/Clover/app/src/main/AndroidManifest.xml
+++ b/Clover/app/src/main/AndroidManifest.xml
@@ -31,7 +31,24 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/Chan.Theme">
-
+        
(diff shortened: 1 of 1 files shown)


Floens:
Is this really needed? Android N has native support for multi-window, and you don't have to do anything for that.

That, and you say you didn't test this, I would appreciate it if I know if this works.

braeden123:
I think it would be useful especially since many devices will never get android N (given android fragmentation/carrier updates). And since a large amount of current devices (most recent LG and Samsung phones) have this capability, there is no harm in providing the option to those users.

How should I go about testing? Clone the repo.. can I compile it into a apk with Android Studio? Not quite familiar with android development. Thanks!

Floens:
Thanks for testing. I don't really like it but it can't hurt.

PR #204, opened 2016-06-26
Use numeric keyboard for legacy CAPTCHA

Florin9doi:


Floens:
I didn't do this because the captcha might change to non numeric numbers and you can't enter the characters anymore with this keyboard.
In fact, you sometimes get a way harder captcha using the legacy captcha that requires non-numeric characters.

Thank you for your contribution anyway.

PR #205, opened 2016-06-26
Change the captcha keyboard to textVisiblePassword

andyklimczak:
This is an advantage because it replaces the autocorrect area of the keyboard with
numbers (which is useful for number only captchas)

Shamelessly stealing idea from thread. Feeling obliged to change it because I changed the keyboard earlier from autocorrect -> non autocorrect, and this is a much better idea than mine.


Floens:
Works great, I didn't know about this input type.

PR #206, opened 2016-06-28
Added zoom to static GiFs

iamzim101:
Had to convert to bitmap myself as the library when given a GIF url, would always fail.


Floens:
The code is not properly formatted, the default formatting of Android Studio should fix it. You changed only the whitespace of CustomScaleImageView.java.

Nitpicking aside, we have to be very careful to not take up too much ram, and that's exactly what SubsamplingScaleImageView is good at, it decodes regions with a lower resolution when zoomed and dynamically loads higher res versions the more you zoom in.

Unfortunately the region decoder doesn't support gifs, but the normal decoder does. Subsampling scale image view lets use disable tiling that that's what we should do.

PR #208, opened 2016-07-04
Change the repliesTo for post from list -> set.

andyklimczak:
This removes the bug when a post seems to have more replies than it should due to people quoting the post number more than once in their post.
This also fixes the bug where the post1 that quotes post2 multiple times, post1 shows up in the list of replies for post2 multiple times.

ie, a single post that looks like:
```javascript
>>3875022
>>3875022
>>3875022
>>3875022
```
Made the post numbered 3875022 think it had 4 replies, when in actuality it was only one.

My tests involved replying to myself from pc -> phone, and phone -> pc and everything matched up correctly. Let me know if you see any other issues or test cases that I might be missing.


Floens:
Thanks. This was messed up after optimizing the logic that handles the replies from/to in chanreader.

PR #209, opened 2016-07-04
Remove google play link

andyklimczak:
Based on your last updates, it seems like google play is completely out of the question now.
Fdroid doesn't have rating or anything it seems, so can't replace this with that.


Floens:
Huh, I thought I still had private_strings, well, private. 
You're right, it should be removed. Link was permanently broken anyway (We can never get that google play package name back again)

PR #211, opened 2016-07-09
Setting to save images into a folder named the current board

iamzim101:
When I formatted before commiting android studio moved MAX_RENAME_TRIES and ImageSaver GetInstance() to where they are now.


Floens:
A few things, you named the var `Board_Name` but the standard for Java is `boardName`.
The ImageSaver is a singleton, it is not save to add mutable fields like you did. It's better to pass the option to the ImageSaveTask.

iamzim101:
Fixed it up with your standards

Floens:
I fixed up the PR some, see the commit message. https://github.com/Floens/Clover/commit/1510b1276614b496ffca747ce1122e43f4508956
I also moved the setting to the main settings screen.
Thanks!

astatine:
This is great; I was about to submit a Feature Request for the same. I have one more addition to the same FR: provide the option to _selectively_ add `.nomedia` files to these subfolders (of our choice; so yes for media from `/b/` but not for media from `/sci/`).

I guess it would be no big deal to do it manually, but an option to do it from within the app would be great.

PR #219, opened 2016-08-06
Text only mode

andyklimczak:
Added to 'Browsing' group in settings to make Clover text only/hide all images. This hides images in Board view, and inside threads. Images are still parsed and can be seen by opening gallery.

Wasn't too sure about the wording. I think I like 'Text only' better than 'Hide images'. But it could go either way.

This idea came up in one of the threads (with quite a few positive replies), and I thought it was a good idea too (for browsing in public reasons)


Floens:
Looks good so far.
Do you think we need to hide images in the catalog mode too? Maybe it looks weird if you hide them there.
Also, I think a reload is needed when you change the setting, so a `RefreshUIMessage` message needs to be broadcast.

andyklimczak:
Setting has catalog mode hiding images, and changing the setting does a RefreshUIMessage.

I think changing both modes with this setting is a good idea, and it does look a little strange (because you're expecting images)

Floens:
Yup looks good to me, thanks as always!

Issue #218, opened 2016-08-01
Unintuitive behavior with back button and cross thread links


finnconor:
Unintuitive behavior with back button and cross thread links

Not a bug, rather merely unexpected or unintuitive behaviour. Press on a cross thread link, accept the warning that you are about to switch thread. Then press back to return to the initial thread and instead of going back to the initial thread you are brought back to the catalogue. This seems unintuitive to me.

Clover version: v2.1.3

PR #223, opened 2016-09-04
Added proper back-press behaviour for cross-thread links

Hual:
Now goes back to the thread where the link was opened.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ui/controller/ViewThreadController.java
+++ b/Clover/app/src/main/java/org/floens/chan/ui/controller/ViewThreadController.java
@@ -30,6 +30,7 @@ import org.floens.chan.controller.NavigationController;
 import org.floens.chan.core.manager.WatchManager;
 import org.floens.chan.core.model.Loadable;
 import org.floens.chan.core.model.Pin;
+import org.floens.chan.core.pool.ThreadFollowerPool;
 import org.floens.chan.core.presenter.ThreadPresenter;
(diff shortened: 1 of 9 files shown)


beschoenen:
Will this also exit Clover when I opened a thread from say Chrome?

Hual:
No, it only follows in-app links.

Floens:
You forgot to commit all files, the CI check failed.

Hual:
I'm an idiot. Added the missing file and CI build is successful now.

Floens:
This needs to be clean up / looked at first. There are some formatting errors and I feel we need a bit of refactoring first.
This PR is related to #218

Clover/app/src/main/java/org/floens/chan/ui/controller/ViewThreadController.java
```diff
@@ -258,4 +271,21 @@ private void setPinIconState() {
...
+
+    @Override
+    public boolean threadBackPressed() {
+        if(threadFollowerPool.empty()) {
+            return false;
+        } else {
+            Loadable threadLoadable = threadFollowerPool.get();
+            loadThread(threadLoadable, false);
+
+            return true;
```
andyklimczak: can be simplified slightly:
```java
    @Override
    public boolean threadBackPressed() {
        if(threadFollowerPool.empty()) {
            return false;
        }
        Loadable threadLoadable = threadFollowerPool.get();
        loadThread(threadLoadable, false);
        return true;
    }
```

PR #225, opened 2016-09-05
Change doneDrawable icon from (CHECK) to (X)

andyklimczak:
This changes the icon for Close when looking at Post Replies from a check to X.

I think it makes a little more sense with an x, but I understand how the check would make sense when calling the variable _done_ drawable.

If you don't think so, then no worries :rocket:


Floens:
This has been toggled multiple times in the three years Clover has existed. I don't know if this is better.

andyklimczak:
That's funny. It's just subjective then

PR #229, opened 2016-09-11
Add images to ImageViewer while it is open

beschoenen:
This PR adds the ability to add images to the thread ImageViewer while it is open.

It is kinda hacky atm and I would like your feedback on it, @Floens.

Please don't merge at this time!


Clover/app/src/main/java/org/floens/chan/core/presenter/ImageViewerPresenter.java
```diff
@@ -52,18 +52,29 @@ public ImageViewerPresenter(Callback callback) {
     }
 
     public void showImages(List<PostImage> images, int position, Loadable loadable) {
-        this.images = images;
-        selectedPosition = position;
+        MultiImageView.Mode newMode;
+        if(this.images.size() < 1) {
```
Floens: Use `list.isEmpty()` instead of `.size() < 1`.

Clover/app/src/main/java/org/floens/chan/core/presenter/ImageViewerPresenter.java
```diff
@@ -52,18 +52,29 @@ public ImageViewerPresenter(Callback callback) {
...
+        }
+
+        this.images.addAll(images);
+        this.selectedPosition = position;
         this.loadable = loadable;
 
-        progress = new ArrayList<>(images.size());
-        for (int i = 0; i < images.size(); i++) {
-            progress.add(i, -1f);
+        this.progress = new ArrayList<>(this.images.size());
```
Floens: `this.progress` can be `progress`.

Clover/app/src/main/java/org/floens/chan/core/presenter/ImageViewerPresenter.java
```diff
@@ -52,18 +52,29 @@ public ImageViewerPresenter(Callback callback) {
     }
 
     public void showImages(List<PostImage> images, int position, Loadable loadable) {
-        this.images = images;
-        selectedPosition = position;
+        MultiImageView.Mode newMode;
+        if(this.images.size() < 1) {
+            newMode = MultiImageView.Mode.LOWRES;
```
Floens: Why is this needed?

Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
```diff
@@ -250,22 +250,36 @@ public void onChanLoaderData(ChanThread result) {
...
             List<Post> posts = result.posts;
+            List<PostImage> images = new ArrayList<>();
+
             int more = 0;
             if (lastLoaded > 0) {
                 for (int i = 0; i < posts.size(); i++) {
                     Post post = posts.get(i);
+
+                    if(newPosts) {
+                        if(post.hasImage) {
```
Floens: Please collapse these two `if`s

Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
```diff
@@ -265,12 +265,17 @@ public void setPagerVisiblity(boolean visible) {
         pager.setSwipingEnabled(visible);
     }
 
+    private ImageViewerAdapter adapter;
```
Floens: Members up top.

Floens:
Hey, a few remarks:
You committed a package change too, I don't think that was your intention.
Please cohere to the default formatter of Android Studio, I see some missing spaces etc.

I added some comments to the commits. Please don't use `this.` when it's not needed.

Maybe it's cleaner to have only one setter, instead of a setter in the contructor and an adder method. We could have a check to see if the image viewer is not already open to open it, and then always set the image list to the image viewer. This also eliminates the need of tracking what new images there are, you just set the whole image list to the image viewer.

You always add images if there are new posts with images, but the image viewer doesn't always show the images of the thread. When you have a reply popup open then the image viewer will only show the images from the popup.

And a final remark, I think `imageViewerNavigationController` is leaked in the `ThreadController`. Once you have opened an image it will stay there until you reopen the viewer. It should either nullify the field when the imageviewer is closed, or (imo better) look if `presentingController` is a `ImageViewerNatigationController` and then get a reference that to that, and then set new images onto that!

beschoenen:
>You committed a package change too, I don't think that was your intention.

I did that so I wouldn't overwrite my current (working) Clover. I'll change it back later on.

>Please cohere to the default formatter of Android Studio, I see some missing spaces etc.

Not sure where you see those

>And a final remark, I think imageViewerNavigationController is leaked in the ThreadController. ...

```Java
imageViewerNavigationController = new ImageViewerNavigationController(context) {
    @Override
    public boolean onBack() {
        imageViewerNavigationController = null;
        return super.onBack();
    }
};
```
This probably isn't what you had in mind, right?

Issue #239, opened 2016-10-05
Changing the save location crashes.


Floens:
Changing the save location crashes the app.

Install a fresh copy of Clover, go to the settings and change the save directory to reproduce the issue.

Android version: 7.0
Phone model: Nexus 5X
Clover version: v2.2.0

bleakoutside:
Same for me. Downgrading to the prevous version, changing the save location and then upgrading retains the save location.

Android version 6.0
Phone Moto g 2015
Clover version 2.2.0

bleakoutside:
I was incorrect. I forgot I used a tasker script to automove any files in the directory.

bleakoutside:
My tasker workaround was unnecesary. If you save a pic in the default location you can then change the save directory without a crash. Sorry for spammig your thread floens. Thanks for a great app.

Tronmech:
Missed this one...  It appears that the dialog doesn't let you get out of /SD card (aka /storage/emulated...) to the actual physical SD card.

joemck:
Same here. I installed the app on a new phone (Galaxy S7, Marshmallow). Clicking on change save location crashes the app, seemingly because /storage/emulated/0/Clover doesn't exist. Once I created the folder in a file manager, it now lets me browse for a folder.

I'm still stuck inside device storage and can't browse to my SD, but that's a different ticket.

PR #269, opened 2016-12-25
Fixed bug #239 (probably)

shinmai-rookie:
Fixed a bug that made the app crash when the user tried to change the destination folder, and the previous folder didn't exist, which caused an exception that went uncaught. What I made was configuring the FileWatcher utility so that if the folder to open doesn't exist, it shows a Toast message and tries to open the parent folder (recursively, until one is found; if none are found, the same exception is thrown); this is arguably not the best solution, and has a possible problem: every Toast message warns about one folder not existing, so if there are many of those folders, the messages may get a bit confusing. However, it works, and does a better job at warning the user the folder doesn't exist than simply crashing does.
As this is my first pull request, and my first attempt at programming anything in Android, I wait your review, Floens; hopefully, I haven't messed up anything.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/saver/FileWatcher.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/saver/FileWatcher.java
@@ -19,15 +19,19 @@ package org.floens.chan.core.saver;
 
 import android.os.FileObserver;
 import android.util.Log;
+import android.widget.Toast;
 
 import org.floens.chan.core.model.FileItem;
 import org.floens.chan.core.model.FileItems;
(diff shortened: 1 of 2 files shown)


Floens:
This probably fixes the bug.
It's however better to prevent the exception, it's there to indicate that the function was incorrectly called.
Having ui code like this (showing a toast) is not recommended in "core" code.

shinmai-rookie:
Fair enough. I guess the best solution would be finding the code of the button to edit the path, and handling the exception there by making the explorer open the parent path instead, but unfortunately I was unable to find where that code is.

PR #271, opened 2016-12-27
Truncate long posts with ellipsis

andyklimczak:
- Factor the truncate long posts into its own method.
- Add an ellipsis to end of long post text to hint at the user know that there is more text to the post

Comparison (look at sticky ending)

Before:
before

After:
after


Floens:
Neat, thanks!

PR #273, opened 2017-01-01
4chan Pass Icon

andyklimczak:
Added 4chan Pass icon to post
Advanced option to disable

The icon lives next the other icons (locked, sticky, etc)
The real value of since4pass is the year that user bought the pass (instead of just a boolean). The is the value given by the json. But I'm not sure if we should show the year, or how.

Example screen:
screenshot_2016-12-31-21-49-59


mkody:
It's kinda weird to have the icon alone on a single line for every posts that have `since4pass`.
So yeah, finding a way to display the year could make it a bit better.

And :+1: for the boolean in the settings.

Floens:
The icons part of a post already has support for text for the countries. We could refactor this so that every icon can have it's on piece of text. I think having text is good, because the icon already takes up the vertical space, so taking up the empty horizontal space isn't bad.
Personally I don't like this feature of 4chan at all, but I'll let the community decide if we show the icon by default or not. It does take up extra space for the nerds that take the time to show off their pass, and these people are already annoying enough ;)

On another note, the advanced settings screen is getting very cluttered, so a refactoring of the settings sections is needed soon.

andyklimczak:
> The icons part of a post already has support for text for the countries. We could refactor this so that every icon can have it's on piece of text.

Good point, I completely missed this.

> Personally I don't like this feature of 4chan at all, but I'll let the community decide if we show the icon by default or not.

I'm with you there on both (I always skip it/don't notice). I like the option of it being there to match the site, and can be useful (such as for vip, I guess?).

I'm fine with working on this bit, but we can scrap it if we don't want this icon in clover

> On another note, the advanced settings screen is getting very cluttered, so a refactoring of the settings sections is needed soon.

Definitely

Floens:
Yes, this can still be added to Clover, if we do the text thing.

Floens:
I'm working on adding generic icons to the Post object, for multi-site support. Work on this is on the multisite branch.

andyklimczak:
👍

Floens:
https://github.com/Floens/Clover/commit/7dc977eedf3b06d71a8a7790d3713c417639751f
No option yet though, will come later when per-site options are figured out.

andyklimczak:
Nice! let's close this pr then

PR #275, opened 2017-01-03
Version name

andyklimczak:
Original PR #226 

I noticed when retesting this post-v2.2.2 release that the command, `git describe --tags --always`, gives the incorrect version number (v2.1.3). That is because it seems v2.2.2 was not tagged to a commit.

Output of of all the tags:
```
clover · dev ⟩ git tag
v1.0
v1.0.1
v1.0.2
v1.1
v1.1.1
v1.1.2
v1.1.3
v1.2.0
v1.2.1
v1.2.10
v1.2.11
v1.2.2
v1.2.3
v1.2.4
v1.2.5
v1.2.6
v1.2.7
v1.2.8
v1.2.9
v2.1.2
v2.1.3
```

### DIFF
--- a/Clover/app/build.gradle
+++ b/Clover/app/build.gradle
@@ -6,7 +6,7 @@ apply plugin: 'com.android.application'
 def getVersionName = { ->
     def stdout = new ByteArrayOutputStream()
     exec {
-        commandLine 'git', 'describe', '--tags'
+        commandLine 'git', 'describe', '--tags', '--always'
         standardOutput = stdout
     }
     return stdout.toString().trim()
--- a/Clover/app/build.gradle
+++ b/Clover/app/build.gradle
@@ -1,5 +1,17 @@
 apply plugin: 'com.android.application'
 
+/**
+ * Gets the version name from the latest Git tag
+ */
+def getVersionName = { ->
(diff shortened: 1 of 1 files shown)


Floens:
I have those versions, try doing a `git fetch --tags` F-Droid requires all versions to have a tag for their update checking.
Now that I think about it, this probably breaks F-Droid's auto update mode. We use the [Tags](https://f-droid.org/manual/fdroid.html#Update-Check-Mode) version checking, see [this](https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/org.floens.chan.txt#L180) and I'm not sure if the F-Droid server can handle this. It doesn't say in the documentation but the F-Droid server looks at the gradle file for the line `versionName` and matches that with the git tag if I remember correctly.

andyklimczak:
You are absolutely correct about fdroid not supporting dynamic versionName. [see first reply here](https://gitlab.com/fdroid/fdroidserver/issues/157)

Instead I have the HEAD commit hash appended to the versionName for debug builds, which is the less fancy method to solve the original problem of 'what debug version am I running?'

Floens:
Yes this looks good, sorry for all the trouble, thanks!

PR #292, opened 2017-03-07
Gradle 2.2.3 -> 2.3.0

andyklimczak:
Build tools 24.0.3 -> 25.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ android:
     - tools
     - extra-android-m2repository
     - build-tools-25.0.0
-    - android-24
+    - android-25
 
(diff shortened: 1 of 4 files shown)


Floens:
`compileSdkVersion` should be upgraded to 25 too. Check if there are any breaking API changes.

andyklimczak:
api changes: https://developer.android.com/sdk/api_diff/25/changes.html

Will research and test more tonight

andyklimczak:
updated and surface tests passed (posting, replying, saving images, gallery, changing save location)

don't feel like this PR is a priority. the only reason I'm aware of gradle updates is because android studio throws popups at me

Floens:
lgtm, thanks for the testing!

PR #297, opened 2017-03-21
Refactor detectLinks

### LINKED ISSUE #278: Can't open URL's surrounded by braces
detailed explanation of the issue]
If there is a URL surrounded with braces, it can not be opened in a browser.
Example:
(https://google.com)
If I click this URL, clover asks me 'Open link'. If I now click OK, I get the message 'No applications found to open link'.

[steps to reproduce the issue]
Write a new post, that contains a link surrounded by braces. Click that link. 

Android version: 6.0.1
Phone model: A0001 (Oneplus One)
Clover version: v2.2.0

andyklimczak:
This replaces the link detection with [autolink-java](https://github.com/robinst/autolink-java#url-extraction) to detect the links in post text.

Done to fix #278 primarily, but will cover a lot more cases now too

Not sure how you feel about adding dependencies to Clover (autolink-java has no dependencies itself)

### DIFF
--- a/Clover/app/build.gradle
+++ b/Clover/app/build.gradle
@@ -130,4 +130,5 @@ dependencies {
     compile 'com.davemorrissey.labs:subsampling-scale-image-view:3.5.0'
     compile 'com.squareup.okhttp3:okhttp:3.4.1'
     compile 'de.greenrobot:eventbus:2.4.0'
+    compile 'org.nibor.autolink:autolink:0.6.0'
 }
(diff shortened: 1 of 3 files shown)


Floens:
Thanks, this is a good addition to the hacky linker we have now.
I try to keep the dependencies down as much as possible. This library is small and does the job well (and only does this job) so I think it's all good.

Quick question before I merge, does this work when the lines start with `>`? For example does `>https://example.com` resolve?

And lastly, the license of the library needs to be added to the licenses.html file, located in the assets/html folder.

andyklimczak:
yes, it covers the ```>``` case:
screenshot_2017-03-21-11-59-32

I also added tests for our use case to robinst/autolink-java#11

and I added the license

Floens:
You've outdone yourself with the tests, merging, thanks!

PR #303, opened 2017-04-12
Add Insomnia theme

ente3000:
I'm actually kind of surprised there wasn't a single pull request for a theme yet.

Preview:

|bjndgwg|9arjz0v|
|:-:|:-:|

I got the color palette from [here](https://userstyles.org/styles/58791/4chan-insomnia) and really liked it, so I ported it.

As the page above states, the CSS code is copyrighted but I'll just rely on these [two](http://graphicdesign.stackexchange.com/questions/43290/copyright-of-a-set-of-colors) [pages](https://www.quora.com/Can-color-palettes-be-copyrighted) and say that the color palette isn't.

Of course I could be wrong, so if I am – sorry.


Floens:
Thanks for a theme, finally someone who adds one!
The copyright is no issue, if there's a complaint we can always remove the theme again.

But before I merge, I think the the contrast of the secondary/hint text is too low. This is the text for the date and no. etc.

ente3000:
You mean the ````post_details_color````? Yeah, I fixed that now.
I also fixed the highlighted post background and dropdown colors.

Before:
screenshot_20170416-165259

After:
screenshot 10

Different resolutions because one screenshot was taken on my phone, the other one with the Android Studio emulator.

Floens:
Thanks!

ente3000:
Great! Thank you for merging it.

PR #309, opened 2017-04-24
Norwegian Bokmål translation

comradekingu:


Floens:
Thank you for your interest in Clover. I am a bit hesitant adding translations to Clover for a few reasons. The strings file is updated regularly, and I don't have the guarantee that I can get a correct translations for it on time. I extremely dislike it when apps are not completely translated. It's also very user unfriendly to have partial translations, because the only way the user can force English is to set the phone system language to English.

Again, I appreciate the work, but I can't merge this right now.

comradekingu:
@Floens Thank you for responding. I share your concern.
One way would be to only have language selectable, and not per default.
Another, and better way, is to register the app on https://hosted.weblate.org/, where more people are able to participate. It is also a good way to spread some awareness.

Floens:
Good news! I've just set up a [crowdin page](https://crowdin.com/project/clover) for helping translating Clover.
I'll try to import the translations you already made.

comradekingu:
@Floens Bad news, crowdin is awful, and closed source adding insult to injury.
Use https://hosted.weblate.org/ or Pootle.

Floens:
I know, but I've made the decision to use crowdin for now.

comradekingu:
@Floens That is a bad decision. Quality suffers as a result, it doesn't have the user-base, and it spies on the user.

Floens:
To take up your suggestion in a more serious sense:

Crowdin does a few things for me:
It is hosted so it does not require any effort on my part hosting wise.
It integrates with GitHub, where it scans the strings.xml file for changes, and updates their internal database with the new strings. This also works the other way around, it automatically creates a new pull request when translations are updated. This lowers the burden of uploading and downloading xml files.
I did take a quick look at weblate and pootle if they had this but I don't think it does, I might be wrong here.

Crowdin is closed-source, but I'm not a software freedom purist, and I don't think this is necessarily evil.

You said crowdin is awful. Could you explain what makes it awful, in contrast to the other services you mentioned?

Are you ok with me uploading your translations to crowdin? I spent the trouble of setting up a translation service for Clover, and the first reaction I get is this negative.

comradekingu:
I am very positive about you doing something about translation, but Crowdin is not the place.

Hosted weblate is hosted, and [integrates with GitHub](https://docs.weblate.org/en/latest/vcs.html#github), same thing.
Hosted Pootle is down right now, but alas, Weblate is good.

Nor am I, and here we are on GitHub, meanwhile for translation the alternatives are better.
Same for same, libre software is a benefit, so Crowdin is already behind there.

For small projects, and even the big ones on there, for branches that have few people in them, strings are essentially locked till the third person comes along, which rarely happens, and takes a lot of time if so.
First person puts in a translation, person two votes against it, you need person 3 to vote for the right thing to make a change. This sounds good, but it isn't how to ensure quality.
Crowdin lacks basic overview of what translations of other strings are, jeopardizing consistency. Sending out error-reports is not automated, etc etc.

comradekingu:
@Floens "Last activity: 6 months ago". Can we reconsidering how good Crowdin is?

4 languages with coverage, in 3 _years_. Don't know how much of that was imported into Crowdin in the first place. Actually -1 language with coverage because nb_NO can't be merged.

The integration doesn't seem to work, as only es-ES, in-ID, ru-rRU and nl-NL exist, and those are manually updated. So one language possibly has coverage, with 0 reviews.
You take donations, and Crowdin only offers hosting for "non-commercial" open source projects, in the monetary sense.

https://hosted.weblate.org/hosting/
https://docs.weblate.org/no/latest/admin/continuous.html?highlight=webhooks#automatically-receiving-changes-from-github
I can help set up shop at Hosted Weblate if you close down Crowdin.

floens:
What a whiny response on a very old PR. If you had looked closely you would have seen that the whole project is at a standstill and that has nothing to do with how the current translation platform works.

PR #322, opened 2017-06-18
Use FileProvider to open videos in external apps

### LINKED ISSUE #310: Can't Use External Player to View WEBMs
I cannot use MX Player or any other external player for viewing WEBMs

I go to open a WEBM file and Clover crashes. I have the external player setting on and I've tried switching that feature on and off multiple times.

Android version: Android 7.0 (literally updated to this not two days ago)
Phone model: SM-920T (Samsung Galaxy Note5)
Clover version: v2.3.0 F-Droid

niririke:
Fixes #310 

If you don't mind.
Feel free to change authority (in manifest and MultiImageView) and names in file_paths. Docs [recommend](https://developer.android.com/reference/android/support/v4/content/FileProvider.html#ProviderDefinition) using my.domain+".fileprovider" (i.e. org.floens.fileprovider) as authority though.

### DIFF
--- a/Clover/app/src/main/AndroidManifest.xml
+++ b/Clover/app/src/main/AndroidManifest.xml
@@ -94,6 +94,18 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
         </receiver>
 
+        <provider
+            android:name="android.support.v4.content.FileProvider"
+            android:authorities="org.floens.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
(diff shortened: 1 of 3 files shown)


Floens:
I'll look at this for the next release.

Floens:
Again, sorry for the delay. I've added a quick workaround for now. I think it's best to do this properly with a content provider when also fixing the sd card issues which also need a proper content provider.

Floens:
Manually merged, thanks 👍

Floens:
I also changed the authorities to change with the package name, so that we can still have both Clover and Clover dev installed at the same time. https://github.com/Floens/Clover/commit/5469528e4bd9880d7b27dc6818b3087c2420316e

PR #338, opened 2017-07-22
Only include text when selecting 'Quote Text'

### LINKED ISSUE #334: For quoting with text, skip post quoting.
For quoting with text, skip post quoting.

When using the quote with text feature, the `>>1234567` quotes are also copied. Stop doing this, because this is almost always not what you want.

Clover version: v2.3.1

andyklimczak:
Do not include `>>post no` or `>>post no (OP)` when quoting with text

Fixes #334

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/presenter/ReplyPresenter.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/presenter/ReplyPresenter.java
@@ -280,8 +280,11 @@ public class ReplyPresenter implements ReplyManager.HttpCallback<ReplyHttpCall>,
 
         if (withText) {
             String[] lines = post.comment.toString().split("\n+");
+            final Pattern quotePattern = Pattern.compile("^>>\\d+(\\s\\(OP\\))?$"); // matches for >>123 or >>123 (OP)
             for (String line : lines) {
-                textToInsert += ">" + line + "\n";
+                if(!quotePattern.matcher(line).matches()) { // do not include post no from quoted post
+                    textToInsert += ">" + line + "\n";
+                }
             }
         }
 


Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
```diff
@@ -55,8 +55,8 @@
 import static org.floens.chan.utils.AndroidUtils.getString;
 
 public class ThreadPresenter implements ChanLoader.ChanLoaderCallback, PostAdapter.PostAdapterCallback, PostCellInterface.PostCellCallback, ThreadStatusCell.Callback, ThreadListLayout.ThreadListLayoutPresenterCallback {
-    private static final int POST_OPTION_QUOTE = 0;
-    private static final int POST_OPTION_QUOTE_TEXT = 1;
+    public static final int POST_OPTION_QUOTE = 0;
+    public static final int POST_OPTION_QUOTE_TEXT = 1;
```
andyklimczak: not sure if I should have made them all public to be consistent, so I only changed what I needed

Floens: I'm not sure either, let's keep it this way.

Floens:
It was unclear in the issue I created, sorry about that.
What I meant was that when you use the quote text feature you *do* also copy the post number from the post you reply to along with the text in the post. But, instead of taking all text in the post and putting `>` before each line, skip lines in the text that match the regex `^>>\d+$`.

So again, quote: take the post's no and insert `>>POSTNO` in the textfield.
Quote text: take the post's no and insert `>>POSTNO` in the textfield, plus any text from the post without lines that match `^>>\d+$`, and put `>` before each line.

The idea behind this change is that you want to quote the text from the post you're replying to to respond to each line. You don't really want to quote the post number quotes of the post you're quoting, because you're not responding to numbers.

Does this sound good to you?

andyklimczak:
I definitely misunderstood, and that sounds good to me

andyklimczak:
Updated, this change was much simpler. My bad about misunderstanding!

andyklimczak:
I realized this would miss `>>123 (You)`

I'll submit a fix, I think there is also another case that I'm missing too

andyklimczak:
cross thread or board linking will also get missed `>>/g/123`

Floens:
I suggest to change the regex to skip anything _starting_ with `>>123 ` and forgetting all the other edge cases. The user can remove any text left it they want.

PR #350, opened 2017-08-20
Simplify regex when quoting with text

andyklimczak:
Simplify regex for removing post no when quoting with text

Matches for everything that it should, but not as performant as it could be, but the `lines` being tested are not that long


andyklimczak:
Not sure why travis failed. I ran the travis script locally and it passed. I bet if we re-run it should be fine

Floens:
Nice regex, looks good to me.

I don't know why the ci fails, travis is killing the process, but like you said nothing has changed.

PR #351, opened 2017-09-04
[WIP] Add parsing to links to provide more context

andyklimczak:
* If the [noEmbed](https://noembed.com/) service can recognize a link when parsing post text, call the noEmbed service and replace the URL text in the post with more helpful text that shows the provider (source) and the title of the link.
* If the noEmbed service doesn't recognize the link, just skip over it and show the link as we do today
* If the noEmbed cannot parse the URL (not a real URL), `noEmbedBuilder.getData()` will throw an exception that we catch, and then show the link as we do today
* URLs will be replaced with the following format: `[Provider] Title of thing` (ie: `[YouTube] Video Title Here`)

I found this no embed service that seemed super useful. Just give it a url and it will spit meta info back about whatever was linked. So I wrote a [small wrapper around it](https://github.com/andyklimczak/No-Embed-JVM), and gave it two methods, a method to check that the url can be used with the service (regex function), and a http call that calls the service and actually gets the info.

What we're doing here, is when iterating through the links in a post, is checking to see if noEmbed is compatible with the link (no network call, just a regex). If it is compatible, then we do the network call for the info, and then replace the URL output with some output that provides more context to the link (ie: youtube video title).

Noembed guarantees that `provider` and `title` will be supplied for every link they handle, so we can always use them to give more context to the links.

This is a WIP because I want to test more, and I want to work with the stuff I had to change with `SpannableStringBuilder` to change the text in the `SpannableString` (`SpannableString` text is immutable) (I'm hoping that you might have an idea that I'm just missing to make it really simple)

Screenshot that should make everything clear:
screenshot_20170903-193055

### DIFF
--- a/Clover/app/build.gradle
+++ b/Clover/app/build.gradle
@@ -112,6 +112,10 @@ android {
     sourceSets {
         beta.java.srcDirs = ['src/release/java']
     }
+
+    packagingOptions {
+        pickFirst 'META-INF/LICENSE'
+    }
 }
 
 dependencies {
@@ -131,4 +135,5 @@ dependencies {
     compile 'com.squareup.okhttp3:okhttp:3.4.1'
     compile 'de.greenrobot:eventbus:2.4.0'
     compile 'org.nibor.autolink:autolink:0.6.0'
+    compile 'com.github.andyklimczak:No-Embed-JVM:c72f2490f6'
 }
--- a/Clover/build.gradle
+++ b/Clover/build.gradle
@@ -11,5 +11,6 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        maven { url 'https://jitpack.io' }
     }
 }
--- a/Clover/app/src/main/java/org/floens/chan/chan/ChanParser.java
+++ b/Clover/app/src/main/java/org/floens/chan/chan/ChanParser.java
@@ -20,6 +20,7 @@ package org.floens.chan.chan;
 
 import android.graphics.Typeface;
 import android.text.SpannableString;
+import android.text.SpannableStringBuilder;
 import android.text.TextUtils;
 import android.text.style.BackgroundColorSpan;
 import android.text.style.StrikethroughSpan;
@@ -57,6 +58,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import noEmbed.NoEmbed;
+import noEmbed.NoEmbedBuilder;
+import noEmbed.NoEmbedException;
 
 import static org.floens.chan.utils.AndroidUtils.sp;
 
@@ -221,7 +226,7 @@ public class ChanParser {
     private CharSequence parseNode(Theme theme, Post post, Node node) {
         if (node instanceof TextNode) {
             String text = ((TextNode) node).text();
-            SpannableString spannable = new SpannableString(text);
+            SpannableStringBuilder spannable = new SpannableStringBuilder(text);
 
             detectLinks(theme, post, text, spannable);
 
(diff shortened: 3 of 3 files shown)


Clover/app/build.gradle
```diff
@@ -112,6 +112,10 @@ android {
     sourceSets {
         beta.java.srcDirs = ['src/release/java']
     }
+
+    packagingOptions {
+        pickFirst 'META-INF/LICENSE'
```
andyklimczak: One of my dependencies in my wrapper library (xml.jackson) was giving me a duplicate file error, and this was a work around

Seemed like a pretty common issue on stackoverflow, but want to look at why some more

andyklimczak:
Also, if you think that this isn't worth the extra load time or should be done a different way, definitely feel free to say so 👍

Floens:
Embedding like this would be a great addition.

Calling the noembed service during parsing is a bad idea, we want the parsing to happen as fast as possible. We should look into a way to call the api after the parsing is done, and updating the post text if necessary.

The noembed services provides an endpoint that returns a list of endpoints that are supported. It would be better if this list is loaded once on startup or the first embed check, because this list is going to change in the future.

It seems that noembed doesn't support querying multiple urls at once.

The state of noembed is uncertain, the project seems to be inactive since October 2016. I can't find any usage terms. If Clover users are going to hit this server, I'm not sure if it will stay up.

Would it be a better idea to support oembed directly? This would eliminate a third party. http://oembed.com/ We then do make a whitelist of domains to embed ourselves.

Clover/app/build.gradle
Floens: Yes I've done this before, this is ok.

Clover/build.gradle
```diff
@@ -11,5 +11,6 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        maven { url 'https://jitpack.io' }
```
andyklimczak: `jitpack` builds maven dependencies on the fly (easy for prototyping like this)

andyklimczak:
>Calling the noembed service during parsing is a bad idea, we want the parsing to happen as fast as possible. We should look into a way to call the api after the parsing is done, and updating the post text if necessary.

Yes, we should do something that reparses the text after it is shown so that it's non blocking. (or something like that)

>The noembed services provides an endpoint that returns a list of endpoints that are supported. It would be better if this list is loaded once on startup or the first embed check, because this list is going to change in the future.

True, we should do something more clever and call that endpoint on startup so we have it. Initially I went the lazier way and have the regex [concatenated and hardcoded](https://github.com/andyklimczak/No-Embed-JVM/blob/master/src/main/kotlin/noEmbed/NoEmbedWrapper.kt#L5). This definitely isn't ideal (since the list could change), calling it in a more proper place on startup then parsing the json response is best. (I also learned while doing this that a big regex is more performant than a bunch of smaller regex)

>The state of noembed is uncertain, the project seems to be inactive since October 2016. I can't find any usage terms. If Clover users are going to hit this server, I'm not sure if it will stay up.

Definitely, the only argument I would say that it's a solid service is that ars technica (somewhat popular tech site, gets linked on hacker news all the time) is a site that uses this. I imagine they have decent traffic, but how often they are calling noembed I'm not sure.

>Would it be a better idea to support oembed directly? This would eliminate a third party. http://oembed.com/ We then do make a whitelist of domains to embed ourselves.

Most likely. It would be a challenge to take on every oembed website, so we could focus on youtube (since that's the most popular request). There are also a few, much better libraries than mine for oembed (I was searching noembed initially), so we could utilize something there too maybe.


I'm just glad we're talking about this feature, whichever way we decide to do it. Maybe we should close this and move discussion to an issue?

PR #372, opened 2017-12-23
Webm Image Search

### LINKED ISSUE #370: Image search for webms submits the wrong format
When using the image search feature on a video, it submits the webm file to whatever service is selected. None of the services accept video files, so the image search fails. 

The solution is to submit the video's thumbnail image instead.
A more sophisticated solution could be to upload a still of the current video position.

This problem can be demonstrated by opening a webm post and selecting Image search>iqdb in the overflow menu. The resulting error page shows that the submitted link is to the webm.

v2.3.1

andyklimczak:
- When image searching using a webm, submit the thumbnail instead
- Image searches don't support webm

Fixes #370

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
+++ b/Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
@@ -204,7 +204,8 @@ public class ImageViewerController extends Controller implements ImageViewerPres
                     public void onFloatingMenuItemClicked(FloatingMenu menu, FloatingMenuItem item) {
                         for (ImageSearch imageSearch : ImageSearch.engines) {
                             if (((Integer) item.getId()) == imageSearch.getId()) {
-                                AndroidUtils.openLinkInBrowser((Activity) context, imageSearch.getUrl(presenter.getCurrentPostImage().imageUrl));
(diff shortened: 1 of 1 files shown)


Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
```diff
@@ -534,4 +535,13 @@ private Window getWindow() {
...
         ImageViewerCallback goToPost(PostImage postImage);
     }
+
+    /**
+     * Send thumbnail image of webm posts because none of the image search providers support webm directly
+     * @param postImage
+     * @return url of an image to be searched
+     */
+    private String getSearchImageUrl(final PostImage postImage) {
+        return postImage.extension.matches("^webm$") ? postImage.thumbnailUrl : postImage.imageUrl;
```
Floens: I think `.matches("^webm$")` is the same as `.equals("webm")`.

andyklimczak: I'll do one better, I'll change it to postImage.type == PostImage.Type.MOVIE

andyklimczak: done

Floens:
Thanks 👍

PR #381, opened 2018-01-09
Updated de (German) translation

squeezer11:
Commit made via Stringlate

### DIFF
--- a/Clover/app/src/main/res/values-de/private_strings.xml
+++ b/Clover/app/src/main/res/values-de/private_strings.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string-array name="extra_abouts">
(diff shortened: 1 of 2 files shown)


Floens:
Thank you for your interest in Clover. I am a bit hesitant adding translations to Clover for a few reasons. The strings file is updated regularly, and I don't have the guarantee that I can get a correct translations for it on time. I extremely dislike it when apps are not completely translated. It's also very user unfriendly to have partial translations, because the only way the user can force English is to set the phone system language to English.

Again, I appreciate the work, but I can't merge this right now.

squeezer11:
hi i have seen your comment right now. i have translated your complete app to german language. not only partial. thats why i hope you integrating the german language completely to your app.

Floens:
Good news! I've just set up a [crowdin page](https://crowdin.com/project/clover) for helping translating Clover.
I'll try to import the translations you already made.

PR #389, opened 2018-01-13
Updated de (German) translation

squeezer11:
Commit made via Stringlate


Floens:
Please look at my other comment before submitting more translations.

PR #454, opened 2018-03-18
Site specific image folders

### LINKED ISSUE #444: Separate "save images in board folder" according to server
Currently, image from same board letter from.different image boards are being saved to same folder if this option is enabled. Please try to change the behavior

andyklimczak:
- When saving images in board specific folders, also save under site name
- `site/board/file.jpg`

resolves #444

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
+++ b/Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
@@ -69,6 +69,7 @@ import org.floens.chan.ui.view.TransitionImageView;
 import org.floens.chan.utils.AndroidUtils;
 import org.floens.chan.utils.Logger;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


Clover/app/src/main/java/org/floens/chan/ui/controller/ImageViewerController.java
```diff
@@ -250,7 +251,9 @@ private void saveShare(boolean share, PostImage postImage) {
             ImageSaveTask task = new ImageSaveTask(postImage);
             task.setShare(share);
             if (ChanSettings.saveBoardFolder.get()) {
-                task.setSubFolder(presenter.getLoadable().boardCode);
+                task.setSubFolder(presenter.getLoadable().site.name() +
+                        File.separator +
+                        presenter.getLoadable().boardCode);
```
andyklimczak: I didn't make the site specific folder a separate option. So if you save in board specific folders, it will also always be under a board. Wasn't sure if it should be it's own option or not

Looking at this code now, my PR will put images under `site/board/image.jpg`, does it make more sense to put in the image in `board/site/image.jpg`?

Floens: I think that site/board is fine.

Issue #499, opened 2018-05-16
Add sushigirl.us


lc-guy:
https://sushigirl.us
It's standard vichan despite the custom homepage, so there shouldn't be any problems. 

Is there a way to implement "overboard"-style boards, which combine all threads from every board on the website into one feed? It's /kaitensushi/ on there, for example.

jtalowell:
This was added by #506 and is available in the [dev build](https://floens.github.io/Clover/#dev). At the moment there's no support for boards like /kaitensushi/ so that was left out.

Issue #431, opened 2018-02-24
[site request] arisuchan.jp


aliceisfollowingit:
Hello, I would love support for arisuchan.jp in the future.
Here is the [repo](https://github.com/arisu-dev/arisuchan).
I think if your have any question, you could ask them on the [[meta] board](https://arisuchan.jp/q/).
Thanks :smile:.

Floens:
it's based on vichan, so getting support.

jtalowell:
This was added by #506 and is available in the [dev build](https://floens.github.io/Clover/#dev).

PR #506, opened 2018-05-29
Support for arisuchan.jp and sushigirl.us

jtalowell:
Addressing #431 and #499.

I'm not sure how to handle overboards such as arisuchan's [all](https://arisuchan.jp/all/) and sushichan's [kaitensushi](https://sushigirl.us/kaitensushi/) so they are left out.

While generic vichan support is on the way I thought that this might help ease the wait.


Floens:
The multiboards aren't supported, so it's ok to leave them out.

PR #518, opened 2018-07-03
Rework setup flow

Floens:
Remove the introduction / welcome screens and instead show the normal start up screen but empty. Then guide the user towards the site setup screen.


Floens:
empty

This screen is shown on first startup, when nothing is added yet.

PR #536, opened 2018-08-03
maintenance

TacoTheDank:
de.greenrobot is now org.greenrobot, just so you know


Floens:
I told you last time that the event bus doesn't need upgrading, and that we don't use beta versions of dependencies.
Gradle also can't be upgraded yet.

TacoTheDank:
im sorry, i forgot

PR #548, opened 2018-08-29
add support for 55chan.org

morphors:
hello can you check this site support.

https://55chan.org

This is my first contribution. Let me know if something is not okay.


Floens:
Thank you.
Have you tested that this works?

Issue #437, opened 2018-02-28
Post date missing from info popup


bakugo:
In previous versions, the full date was displayed in the post info popup, but this seems to have been commented out during multisite development and never added back in, so right now it's impossible to see a post's date if relative timestamps are enabled.

Related: Why is the info button hidden in the "More..." menu? Basic post info like the date and filename shouldn't require 2 taps to access imo.

PR #565, opened 2018-10-04
Accessible post-info dialogue option

ahldr:
- Reimplemented post date
 - Added option for a 1-tap post info button, rather than a 2 tap
 button

This is my ~~first~~ second PR to this project so feel free to yell at me if anything's wrong.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
@@ -51,8 +51,10 @@ import org.floens.chan.ui.view.FloatingMenuItem;
 import org.floens.chan.ui.view.ThumbnailView;
 import org.floens.chan.utils.AndroidUtils;
 
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -480,9 +482,16 @@ public class ThreadPresenter implements ChanThreadLoader.ChanLoaderCallback, Pos
             menu.add(new FloatingMenuItem(POST_OPTION_DELETE, R.string.post_delete));
         }
 
+        if(ChanSettings.accessibleInfo.get()){
+            //Accessible info enabled
+            menu.add(new FloatingMenuItem(POST_OPTION_INFO, R.string.post_info));
+        } else {
+            extraMenu.add(new FloatingMenuItem(POST_OPTION_INFO, R.string.post_info));
+        }
+
         menu.add(new FloatingMenuItem(POST_OPTION_EXTRA, R.string.post_more));
 
-        extraMenu.add(new FloatingMenuItem(POST_OPTION_INFO, R.string.post_info));
+
         extraMenu.add(new FloatingMenuItem(POST_OPTION_LINKS, R.string.post_show_links));
         extraMenu.add(new FloatingMenuItem(POST_OPTION_OPEN_BROWSER, R.string.action_open_browser));
         extraMenu.add(new FloatingMenuItem(POST_OPTION_SHARE, R.string.post_share));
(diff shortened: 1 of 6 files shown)


alex-eyre:
Response to #437

Clover/app/src/main/java/org/floens/chan/ui/controller/AppearanceSettingsController.java
```diff
@@ -81,6 +81,10 @@ private void populatePreferences() {
                     R.string.setting_enable_reply_fab,
                     R.string.setting_enable_reply_fab_description)));
 
+            requiresUiRefresh.add(layout.add(new BooleanSettingView(this,
+                    ChanSettings.accessibleInfo,
+                    "Enable accessible post info",
+                    "Place post info in a more accessible location")));
```
andyklimczak: you should put these raw strings in [`strings.xml`](https://github.com/Floens/Clover/blob/3c7f0ddfff53d179703c78248c864ceb68acd67f/Clover/app/src/main/res/values/strings.xml)

Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
```diff
@@ -698,6 +707,9 @@ private void showPostInfo(Post post) {
         }
 
         // TODO(multi-site) get this from the timestamp
+        Date date = new Date () ;
+        date.setTime((long)post.time*1000);
+        text.append(new StringBuilder("Date: " + date.toString()));
```
andyklimczak: instead of creating a new `StringBuilder` object,  you could do something like `text.append("Date: ").append(date.toString()));`

Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
```diff
@@ -698,6 +707,9 @@ private void showPostInfo(Post post) {
         }
 
         // TODO(multi-site) get this from the timestamp
+        Date date = new Date () ;
+        date.setTime((long)post.time*1000);
```
andyklimczak: does this time when outputted respect the user's current timezone, or is it always UTC?

Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
```diff
@@ -51,8 +51,10 @@
 import org.floens.chan.ui.view.ThumbnailView;
 import org.floens.chan.utils.AndroidUtils;
 
+import java.time.Instant;
```
andyklimczak: this import might not be needed

Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java
alex-eyre: It respects the user's current timezone.

alex-eyre:
I'll resolve all these when I get a moment, thank you for the feedback @andyklimczak

Clover/app/src/main/res/values/strings.xml
```diff
@@ -159,7 +159,7 @@ Re-enable this permission in the app settings if you permanently disabled it."</
     <string name="thread_pin_hint">Bookmark this thread</string>
     <string name="thread_view_archive">Archive</string>
 
-    <string name="thread_empty_setup_feature" translatable="false">&#128564;</string>
+    <string name="thread_empty_setup_feature" translatable="false">&#128169;</string>
```
Floens: This seems unnecessary.

floens (changes requested):

Floens:
You seem to have next to setting the date also have added a setting for a non-nested more options setting. This is ok, but I want to briefly explain what I meant with the TODO that was there for the date.
4chan and other sites also have a string-based post date next to the unix timestamp. For example, the 4chan api has the field "now" which contains something like "08\/08\/12(Wed)01:11". It used to work like this before the update.
Maybe you could change the meaning of this PR to a more accessible options menu, and add the functionality described above in another?

Clover/app/src/main/res/values/strings.xml
alex-eyre: I forgot I changed that...
sorry.

Issue #544, opened 2018-08-22
Feature Request: Use ExoPlayer for WebM Playback


Riotous:
Every open source project is moving to or has moved to using Google's Exoplayer for flawless media playback on Android devices no matter what version of Android they run. Clover should follow suit!

PR #566, opened 2018-10-06
Implement ExoPlayer

alex-eyre:
- Add ExoPlayer logic to MultiImageView.java
 - Add option to enable/disable ExoPlayer in MediaSettings (defaults to
 enabled).

Response to #544.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
+++ b/Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
@@ -52,6 +52,7 @@ import org.floens.chan.core.cache.FileCache;
 import org.floens.chan.core.cache.FileCacheDownloader;
 import org.floens.chan.core.cache.FileCacheListener;
 import org.floens.chan.core.cache.FileCacheProvider;
+import org.floens.chan.core.di.UserAgentProvider;
 import org.floens.chan.core.model.PostImage;
 import org.floens.chan.core.settings.ChanSettings;
 import org.floens.chan.utils.AndroidUtils;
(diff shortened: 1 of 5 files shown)


Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
```diff
@@ -451,6 +483,14 @@ private boolean hasMediaPlayerAudioTracks(MediaPlayer mediaPlayer) {
         }
     }
 
+    private boolean hasMediaPlayerAudioTracks(ExoPlayer mediaPlayer) {
+        if(mediaPlayer.getAudioComponent() == null) {
```
Floens: Simplify to `return mediaPlayer.getAudioComponent() != null`.

Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
```diff
@@ -391,6 +408,21 @@ private void setVideoFile(final File file) {
             AndroidUtils.openIntent(intent);
 
             onModeLoaded(Mode.MOVIE, videoView);
+        } else if (ChanSettings.videoUseExoplayer.get()) {
+            exoVideoView = new PlayerView(getContext());
+            exoPlayer = ExoPlayerFactory.newSimpleInstance(getContext());
+            exoVideoView.setPlayer(exoPlayer);
+            DataSource.Factory dataSourceFactory = new DefaultDataSourceFactory(getContext(),
+                    Util.getUserAgent(getContext(),"Clover"));
```
Floens: The `UserAgentProvider` could be used for this. You need to inject it, see other code that uses it.

alex-eyre:
I implemented your suggestions @Floens, I didn't realise my spacings were that bad ¯\_(ツ)_/¯

PR #573, opened 2018-10-17
Added always open links with external browser option

veridiam:
Use cases: ad blocking in mobile firefox, youtube links in youtube app.


Clover/app/src/main/java/org/floens/chan/ui/layout/ThreadLayout.java
```diff
@@ -301,14 +301,22 @@ public void openLink(final String link) {
                     .setPositiveButton(R.string.ok, new DialogInterface.OnClickListener() {
                         @Override
                         public void onClick(DialogInterface dialog, int which) {
-                            AndroidUtils.openLinkInBrowser((Activity) getContext(), link);
+                            if (ChanSettings.openLinkBrowser.get()) {
```
Floens: Could you create a method for this duplicated code?

veridiam:
> Looks good. The code from openLink and openLinkInBrowser is quite complex. This means that the browser is opened instead of the chrome custom tab, right?

Yep, opens in firefox for me

PR #582, opened 2018-11-01
Fix spelling in README.md

Twinov:
spacial > special


Floens:
Thanks. The two spaces at the end are a requirement for correct formatting (forcing newlines for text below it), but at the end of the section it's fine to leave them out. Keep that in mind.

Twinov:
My bad I'm new to GitHub.

PR #584, opened 2018-11-04
Norwegian Bokmål translation

comradekingu:


Floens:
We use crowdin for the translations. We had a previous conversation about this, where you stated that you didn't want your translations in it.

PR #592, opened 2018-11-10
Bug fixes, improvments and 2ch.hk support

softach-corp:
I had to hardcode a few core things in 27f5693, sorry.

If you have better ideas on how to pass a custom cookie to a board request, tell me.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/cache/FileCacheDownloader.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/cache/FileCacheDownloader.java
@@ -24,6 +24,10 @@ import android.support.annotation.MainThread;
 import android.support.annotation.WorkerThread;
 
 import org.floens.chan.core.settings.ChanSettings;
+import org.floens.chan.core.settings.SettingProvider;
+import org.floens.chan.core.settings.SharedPreferencesSettingProvider;
+import org.floens.chan.core.settings.StringSetting;
+import org.floens.chan.utils.AndroidUtils;
 import org.floens.chan.utils.Logger;
 
 import java.io.Closeable;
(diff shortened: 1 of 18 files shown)


Floens:
We can use the intercept infrastructure we have for sites to make this cookie exception. I'll look into it.

Floens:
@softach-corp I'm confused by the nsfw-token stuff.

I'm currently adding the infrastructure to easily modify http calls from FileCache and the proxiedurlstack (volley).

A few questions:
- I don't speak Russian that well :). The nsfw boards are protected (say, /h/), but where do I normally authenticate on the site?
- Could you explain the cookies `ageallow` and `usercode_auth`? Are both needed for nsfw boards?
- The api and images are protected, and that's why volley and the filecache need the cookie header. Do they only need `ageallow=1` for the nsfw boards or also `usercode_auth`?
- Currently the DvachReplyCall utilises the LoginResponse class to feed back the usercode. But this would mean that the token is gathered when posting a reply to a thread. Is this correct?

Basically, I'm confused what the `ageallow` and `usercode_auth` do, where they are required and where we can get them.

Floens:
Is usercode_auth the same as the passcode you can buy at https://2ch.hk/2ch/? (The equivalent of a 4chan pass).
If so, the dvach module should allow the user to login at the dvach settings screen, the same way it is done with 4chan. The logic done now with the replying doesn't work.

softach-corp:
> @softach-corp I'm confused by the nsfw-token stuff.
> 
> I'm currently adding the infrastructure to easily modify http calls from FileCache and the proxiedurlstack (volley).
> 
> A few questions:
> 
> * I don't speak Russian that well :). The nsfw boards are protected (say, /h/), but where do I normally authenticate on the site?
> * Could you explain the cookies `ageallow` and `usercode_auth`? Are both needed for nsfw boards?
> * The api and images are protected, and that's why volley and the filecache need the cookie header. Do they only need `ageallow=1` for the nsfw boards or also `usercode_auth`?
> * Currently the DvachReplyCall utilises the LoginResponse class to feed back the usercode. But this would mean that the token is gathered when posting a reply to a thread. Is this correct?
> 
> Basically, I'm confused what the `ageallow` and `usercode_auth` do, where they are required and where we can get them.

Yes, this is correct. You don't authenticate to get it, if you don't have a usercode token the server will generate it for you when you make any post anywhere. This is to prevent outsiders from nsfw boards (and to prevent abuse reports). No cookie = 404 for all unsafe paths. usercode is only required for specified board list

Actually, is think ageallow is not strictly required, it is only used in browser to check if the age warning banner has been displayed already..

Floens:
Ah I see, so the logic that the cookie is extracted after a post is correct (it seemed so strange).

I'm going to push to your branch soon with fixes for this, and then it's probably ready to merge.

PR #616, opened 2019-01-02
Touch up and complete the current strings.

TacoTheDank:
- Fix and improve some grammar (e.g. some acute accents being used improperly in the Spanish strings, Oxford commas, better sentence structure in the English strings, etc).
- Create the missing ```setting_video_exoplayer``` / ```setting_video_exoplayer_description``` and ```setting_force_english_locale``` / ```setting_force_english_locale_toggle_notice``` string translations that were missing for the other languages. They are accurate for the most part (better than being missing).
- Translate the remaining non-English strings that were still in English. They are accurate for the most part (better than not being translated at all).
- Replace ```&#8230;``` with what it substitutes, which is the singular character ellipses. We can do this because the files support UTF-8.

Signed-off-by: Taco <SkytkRSfan3895@gmail.com>


TacoTheDank:
Likely super outdated, so I will close

PR #629, opened 2019-02-23
#543, add filter masterswitch

### LINKED ISSUE #543: [Request] Filter masterswitch
A main switch for filters would be nice to have, since you may need to disable all of your filters at once in some situations.

Adamantcheese:
Resolves #543

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/ui/controller/FiltersController.java
+++ b/Clover/app/src/main/java/org/floens/chan/ui/controller/FiltersController.java
@@ -50,6 +50,7 @@ import de.greenrobot.event.EventBus;
 
 import static org.floens.chan.Chan.inject;
 import static org.floens.chan.ui.theme.ThemeHelper.theme;
+import static org.floens.chan.utils.AndroidUtils.getAppContext;
 import static org.floens.chan.utils.AndroidUtils.getAttrColor;
 import static org.floens.chan.utils.AndroidUtils.getString;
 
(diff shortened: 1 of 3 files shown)


Adamantcheese:

Clover/app/src/main/java/org/floens/chan/ui/controller/FiltersController.java
```diff
@@ -130,6 +136,33 @@ public void onCreate() {
...
         if (v == add) {
             showFilterDialog(new Filter());
+        } else if (v == enable) {
+            //if every filter is disabled, enable all of them and set the drawable to be an x
+            //if every filter is enabled, disable all of them and set the drawable to be a checkmark
+            //if some filters are enabled, disable them and set the drawable to be a checkmark
+            List<Filter> enabledFilters = filterEngine.getEnabledFilters();
+            List<Filter> allFilters = filterEngine.getAllFilters();
+            if(enabledFilters.isEmpty()) {
+                for (Filter filter : allFilters) {
```
andyklimczak: do you think it would be beneficial to wrap and paramaterize this loop, to reduce duplicate code between the conditionals?
```
if (enabledFilters.isEmpty()) {
  setFiltersEnabled(allFilters, true)
  setDrawableTo(R.drawable.ic_clear_white_24dp, v)
} else if...
```
where `setFilters(List<Filter> filters, boolean enabled)` will set the enabled attribute to the passed in parameter

Adamantcheese: That's a good idea, I'll implement that.

andyklimczak (approved):
LGTM

Issue #484, opened 2018-05-01
Thread title not display correctly when making a thread


c933103:
Step to reproduce:
1. Make a thread
2. It automatically bookmarked
3. The text for bookmarked thread being shown consists of only numbers i.stead of thread title or thread text

Android version: 8.0.0
Phone model: XZ1c
Clover version: 3.0.0

Adamantcheese:
Pull request #632 fixes this issue. Threads aren't loadable as soon as they're posted, so they'll have a default title of the board and OP post number until the thread is loadable. Once it is loaded again by the watch manager it'll update to the proper thread title. It shouldn't be more than 2 or 3 minutes before the thread is loadable and this is the easy, one line of code, eventually correct solution.

Adamantcheese:
I've been hard at work, and submitted 8 pull requests. While Floens and other maintainers work on adding them in, if ever, [you can find a debug multi-release of all the features from my pulls here.](https://github.com/Adamantcheese/Clover/releases)

Issue #450, opened 2018-03-10
After creating a thread, clover might not able to pick up content of thread as summary, instead show thread number in toolbar and title bar


c933103:
![screenshot_20180310-164130](https://user-images.githubusercontent.com/1525293/37240122-ff305e82-2481-11e8-9f16-9892f6cd5261.png)

Android version: 8.0.0
Clover version: 2.3.1-83ae7a29

PR #632, opened 2019-02-28
#484, update pinned thread titles for newly posted threads when possible

Adamantcheese:
I have left an explanation as a comment.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
@@ -761,6 +761,15 @@ public class WatchManager {
         @Override
         public void onChanLoaderData(ChanThread thread) {
             pin.isError = false;
+            /*
+             * Forcibly update watched thread titles
+             * This solves the issue of when you post a thread and the site doesn't have the thread listed yet,
+             * resulting in the thread title being something like /test/918324919 instead of a proper title
+             *
+             * The thread title will be updated as soon as the site has the thread listed in the thread directory
+             * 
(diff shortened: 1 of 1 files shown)


Adamantcheese:

Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
```diff
@@ -761,6 +761,15 @@ public void onChanLoaderError(ChanLoaderException error) {
...
             pin.isError = false;
+            /*
+             * Forcibly update watched thread titles
+             * This solves the issue of when you post a thread and the site doesn't have the thread listed yet,
+             * resulting in the thread title being something like /test/918324919 instead of a proper title
+             *
+             * The thread title will be updated as soon as the site has the thread listed in the thread directory
+             * 
+             */
+            pin.loadable.setTitle(PostHelper.getTitle(thread.op, pin.loadable));
```
andyklimczak, line 772: not sure how expensive `.setTitle` is, but maybe we should `.setTitle` only if the current `pin.loadable.title` does not equal `PostHelper.getTitle(thread.op, pin.loadable)`. something like:
```
if (!pin.loadable.getTitle().equals(PostHelper.getTitle(thread.op, pin.loadable)) {
  pin.loadable.setTitle(PostHelper.getTitle(thread.op, pin.loadable));
}
```

minor thing really I think, so not needed

andyklimczak (approved):
LGTM

Adamantcheese:
This might also be a solution to #450, but it might not.

PR #633, opened 2019-03-01
#252, add page viewing (at least for 4chan)

### LINKED ISSUE #252: [Suggestion] Thread page number
It would be very useful to be able to see what page the thread you're in is currently at, right next to the post number etc.
I frequent a slow general that needs a bump every now and then, but I can never tell when on mobile.

Clover version: 2.2.0

Adamantcheese:
Loads up threads.json from the site every time the thread is loaded or reloaded and puts that info in the images/replies/posters area. If a thread is archived, it'll be '?' instead. Fixes #252


Adamantcheese:

Clover/app/src/main/java/org/floens/chan/core/manager/PageRequestManager.java
```diff
@@ -0,0 +1,153 @@
...
+    @Inject
+    public PageRequestManager() {
+    }
+
+    public Page getPage(Post op) {
+        if (op == null) {
+            return null;
+        }
+        Pages pages = getPages(op.board);
+        if (pages == null) {
```
andyklimczak: could we dedupe the code here?

maybe it could be a method on the `Pages` model? `pages.findPageForThread(ThreadTime threadTime)` to hide this logic away

Clover/app/src/main/java/org/floens/chan/core/manager/PageRequestManager.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+        return null;
+    }
+
+    private Pages getPages(Board b) {
+        if (savedBoards.contains(b.code)) {
+            //if we have it stored already, return the pages for it
+            //also issue a new request if 3 minutes have passed
+            long lastUpdateTime = boardTimeMap.get(b.code);
+            if (lastUpdateTime + 3 * 1000 * 60 > System.currentTimeMillis()) {
```
andyklimczak: you could flip this boolean and the else into the if block, to avoid empty blocks

andyklimczak: maybe constantize the `3 * 1000 * 60` (`THREE_MINUTES`?) and/or wrap this check in a method that could be called  something like `shouldThreadUpdate(lastUpdateTime)`

Clover/app/src/main/java/org/floens/chan/core/site/ThreadTime.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.floens.chan.core.site;
+
+public class ThreadTime {
+    public final int no;
+    public final long modified;
+
+    public ThreadTime(int no, long modified) {
+        this.no = no;
+        this.modified = modified;
```
andyklimczak: Could these attributes live on the `ChanThread` model? (in case you didn't think about it)

EDIT: I see the response `http://a.4cdn.org/g/threads.json`, which drove you to create a specific model for that response. I guess I was trying to see if there was a way to consolidate data and remove either/both `Map`s in `PageRequestManager`

Clover/app/src/main/java/org/floens/chan/ui/cell/ThreadStatusCell.java
```diff
@@ -147,6 +148,20 @@ public boolean update() {
...
+                    if (p != null) {
+                        page = page + String.valueOf(p.page);
+                        pageS = new SpannableString(page);
+                        if (p.page >= board.pages) {
+                            pageS.setSpan(new StyleSpan(Typeface.ITALIC), 0, page.length(), 0);
+                        }
+                    } else {
+                        pageS = new SpannableString(page + "?");
+                    }
+                    finalText = TextUtils.concat(finalText, " / ", pageS);
```
andyklimczak: I know this wasn't you, but it might be good to refactor to build `result` using [`SpannableStringBuilder`](https://developer.android.com/reference/android/text/SpannableStringBuilder.html)

andyklimczak (commented):
There is unfortunate UX where if you open clover, scroll/open any page 3 thread, post something to bump it to page 1, the text at the bottom will still say `Page 3` until the 3 minute timer comes up. Not entirely sure what can be done

It's too bad that the `last_modified` field from the `http://a.4cdn.org/g/threads.json` endpoint isn't on the `http://a.4cdn.org/g/1.json` or `http://a.4cdn.org/g/thread/70313875.json` endpoints so that you could update a single thread as it gets auto updated or when you post. Maybe you could do something by looking at the last post's create time?

Clover/app/src/main/java/org/floens/chan/ui/cell/ThreadStatusCell.java
Adamantcheese: I can do that.

Clover/app/src/main/java/org/floens/chan/core/manager/PageRequestManager.java
Adamantcheese: I can implement both of those.

Clover/app/src/main/java/org/floens/chan/core/manager/PageRequestManager.java
Adamantcheese: Sure, the logic is basically the same.

Clover/app/src/main/java/org/floens/chan/core/site/ThreadTime.java
Adamantcheese: Yeah, the `ChanThread` model is only barely related to the pages model in that the OP number is used to relate them. I can refactor the ThreadTime moniker to something that's a bit less confusing, and put all the page related objects into one spot with a good label.

Adamantcheese:
It's a bit annoying, because like in #632 it's not immediately consistent and trying to force it would be way more work for something that's honestly not that big of a deal. I can force it to update in a time less than 3 minutes without issue. The reason for the three minutes was that otherwise it would be requesting pages every second, which is obviously a terrible idea for something that doesn't update that fast.

EDIT: Also the pages.json doesn't update consistently across boards. /a/'s seems to update every 20 seconds, while /g/ is 16 and /qa/ is immediate. I've set the default re-update after post to 30 seconds. The force update can probably be added elsewhere as well, but it's not necessary.

Issue #425, opened 2018-02-20
8chan: cross-board links apparently doesn't work?


c933103:
https://8ch.net/test/res/60226.html when I click on >>>/test2/ it said "No applications found to open link"

bakugo:
This happens on 4chan too with board links (such as `>>>/b/`), haven't tested with post links.

PR #634, opened 2019-03-01
#599, fixes cross board links, adds structure for direct board links and board search links

### LINKED ISSUE #599: Cross-board links not working on 4chan
Tapping cross-board links such as in the image causes nothing to happen. 
https://i.imgur.com/Xs6B2i2.jpg

Android version: 8.1
Clover version: Both 3.0.1 stable and the most recent dev build

Adamantcheese:
Fixes #599, 4chan specific bug due to domains.
EDIT 5/7/2019: #425 is also solved.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/site/parser/CommentParser.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/site/parser/CommentParser.java
@@ -26,6 +26,7 @@ import android.text.style.UnderlineSpan;
 
 import org.floens.chan.core.model.Post;
 import org.floens.chan.core.model.PostLinkable;
+import org.floens.chan.core.site.Page;
 import org.floens.chan.ui.span.AbsoluteSizeSpanHashed;
 import org.floens.chan.ui.span.ForegroundColorSpanHashed;
 import org.floens.chan.ui.theme.Theme;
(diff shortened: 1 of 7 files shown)


Adamantcheese:

Clover/app/src/main/java/org/floens/chan/core/site/parser/CommentParser.java
```diff
@@ -233,6 +235,12 @@ public CharSequence handleTable(Theme theme,
 
     public Link matchAnchor(Post.Builder post, CharSequence text, Element anchor, PostParser.Callback callback) {
         String href = anchor.attr("href");
+        // gets us something like /board/thread/postno#quoteno
+        //hacky fix for 4chan having two domains but the same API
+        if(href.matches("//boards\\.4chan.*?\\.org/(.*?)/thread/(\\d*?)#p(\\d*)")) {
+            href = href.substring(2);
+            href = href.substring(href.indexOf('/'));
```
andyklimczak, line 243: instead of string manipulation, could you use the [Uri class](https://developer.android.com/reference/android/net/Uri.html)?

Adamantcheese: The rest of the method expects something in the form of `(thread no)#p(post no)` and does regex matching against that; this code gets the string value to the correct form as quickly as possible. It's possible that the entire method may be redone with URI's, however I don't know enough about the architecture of the project to make that call.

Adamantcheese:
Fixes #425 also.

PR #636, opened 2019-03-04
#580, last page notifications

### LINKED ISSUE #580: Notify when thread you are watching reaches the last page
If I'm watching a thread, I usually don't want it to die so it would be nice if I could be notified when the thread is close to being archived. This is especially useful for boards that move fast.

Adamantcheese:
Fixes #580. Relies on the solution done in #252 for some functionality, which is included in this branch.


Adamantcheese:

TacoTheDank:
Nice job on all of these :) I've been getting antsy for a new dev build, so I'll download the apk and try it out.

Adamantcheese:
I've made a fix that makes last page updates quicker, due to an oversight. You can grab that new release at the link above.
I'll take a look, but I give no guarantees. That would fall under the slew of "add an archive" requests.

Adamantcheese:
I'll take a look, but I give no guarantees. That would fall under the slew of "add an archive" requests. Submit a new issue for your feature request.

Issue #607, opened 2018-12-10
Feature request: jump to next (you) 


observeroftime02:
[short and descriptive title]
Requesting the feature to jump to the next post mentioning you

[detailed explanation of the issue]
Sometimes in long threads it can be a challenge to find the posts quoting you. I'd have to scroll through and look for the backlink followed by (you) to identify who's quoting me, or scroll back to find my post and click on the reply count. 

Either a menu option to jump to the next poster quoting me or a real visual highlight cue of posts quoting me would be beneficial for usability.

happapa:
I would prefer a similar hilighting you can already get for posts using filters. Sadly ">>XXXXXXX (You)" is not part of the message body, so you can't currently add a regex filter for it inside the app.

WaywardSkylark:
A bit of a workaround I use is just searching for (you). It highlights replies well enough.

PR #640, opened 2019-03-09
Added 'Show replies' option to menu

chieres:
Added "Show replies" option which automatically does a search for " (You)".
Initially added because I'd use this often, but now that I've checked issues it's actually related to [Issue 607](https://github.com/Floens/Clover/issues/607).


floens:
Would this be sufficient though? Can't we add a filter to see if `isSaved` is true?

chieres:
> Would this be sufficient though? Can't we add a filter to see if `isSaved` is true?

If by "Sufficient" you mean "if it works" - yes it does, I've checked it on my device.
If by that you mean "is this the most elegant solution" - then probably not. 

I do not know the internals of the app so I've come up with this easy solution at the time. But I can try and dig deeper for a proper solution whenever I have time.

PR #654, opened 2019-03-30
#644, Recursive filtering

### LINKED ISSUE #644: [suggestion] Recursive Filtering
Basically that when you filter someone you hide not only the poster, but also everyone who replies to it, hiding long chains of shitposting.

K1rakishou:
Closes #644.

Added two new menu items - "Hide" and "Hide whole chain". "Hide" is for hiding separate posts/threads whereas "Hide whole chain" is for hiding posts with all replies to those posts. OP post cannot be hidden from a thread (only from the catalog and in this case the whole thread will be hidden).

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
@@ -45,7 +45,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
     private static final String TAG = "DatabaseHelper";
 
     private static final String DATABASE_NAME = "ChanDB";
-    private static final int DATABASE_VERSION = 26;
+    private static final int DATABASE_VERSION = 27;
 
     public Dao<Pin, Integer> pinDao;
     public Dao<Loadable, Integer> loadableDao;
@@ -265,6 +265,17 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
                 postHideDao.executeRawNoArgs("ALTER TABLE " + POST_HIDE_TABLE_NAME + " ADD COLUMN whole_thread INTEGER default 0");
             } catch (SQLException e) {
                 Logger.e(TAG, "Error upgrading to version 26", e);
+            }
+        }
+
+        if (oldVersion < 27) {
+            try {
+                // Create indexes for PostHides to speed up posts filtering
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_site_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.SITE_COLUMN_NAME + ");");
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_board_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.BOARD_COLUMN_NAME + ");");
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_no_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.NO_COLUMN_NAME + ");");
+            } catch (SQLException e) {
+                Logger.e(TAG, "Error upgrading to version 27", e);
             }
         }
     }
--- a/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
@@ -45,7 +45,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
     private static final String TAG = "DatabaseHelper";
 
     private static final String DATABASE_NAME = "ChanDB";
-    private static final int DATABASE_VERSION = 27;
+    private static final int DATABASE_VERSION = 28;
 
     public Dao<Pin, Integer> pinDao;
     public Dao<Loadable, Integer> loadableDao;
@@ -57,6 +57,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
     public Dao<SiteModel, Integer> siteDao;
 
     public static final String POST_HIDE_TABLE_NAME = "posthide";
+    public static final String FILTER_TABLE_NAME = "filter";
 
     private final Context context;
 
@@ -276,6 +277,15 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
                 postHideDao.executeRawNoArgs("CREATE INDEX posthide_no_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.NO_COLUMN_NAME + ");");
             } catch (SQLException e) {
                 Logger.e(TAG, "Error upgrading to version 27", e);
+            }
+        }
+
+        if (oldVersion < 28) {
+            try {
+                postHideDao.executeRawNoArgs("ALTER TABLE " + POST_HIDE_TABLE_NAME + " ADD COLUMN " + PostHide.HIDE_COLUMN + " INTEGER default 0");
+                filterDao.executeRawNoArgs("ALTER TABLE " + FILTER_TABLE_NAME + " ADD COLUMN " + Filter.APPLY_TO_REPLIES_COLUMN + " INTEGER default 0");
+            } catch (SQLException e) {
+                Logger.e(TAG, "Error upgrading to version 28", e);
             }
         }
     }
--- a/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
@@ -283,6 +283,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
         if (oldVersion < 28) {
             try {
                 postHideDao.executeRawNoArgs("ALTER TABLE " + POST_HIDE_TABLE_NAME + " ADD COLUMN " + PostHide.HIDE_COLUMN + " INTEGER default 0");
+                postHideDao.executeRawNoArgs("ALTER TABLE " + POST_HIDE_TABLE_NAME + " ADD COLUMN " + PostHide.HIDE_REPLIES_TO_THIS_POST_COLUMN + " INTEGER default 0");
                 filterDao.executeRawNoArgs("ALTER TABLE " + FILTER_TABLE_NAME + " ADD COLUMN " + Filter.APPLY_TO_REPLIES_COLUMN + " INTEGER default 0");
             } catch (SQLException e) {
                 Logger.e(TAG, "Error upgrading to version 28", e);
--- a/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
@@ -262,8 +262,8 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
 
         if (oldVersion < 26) {
             try {
-                postHideDao.executeRawNoArgs("ALTER TABLE threadhide RENAME TO " + POST_HIDE_TABLE_NAME + ";");
-                postHideDao.executeRawNoArgs("ALTER TABLE " + POST_HIDE_TABLE_NAME + " ADD COLUMN whole_thread INTEGER default 0");
+                postHideDao.executeRawNoArgs("ALTER TABLE threadhide RENAME TO posthide;");
+                postHideDao.executeRawNoArgs("ALTER TABLE posthide ADD COLUMN whole_thread INTEGER default 0");
             } catch (SQLException e) {
                 Logger.e(TAG, "Error upgrading to version 26", e);
             }
@@ -272,9 +272,9 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
         if (oldVersion < 27) {
             try {
                 // Create indexes for PostHides to speed up posts filtering
-                postHideDao.executeRawNoArgs("CREATE INDEX posthide_site_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.SITE_COLUMN_NAME + ");");
-                postHideDao.executeRawNoArgs("CREATE INDEX posthide_board_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.BOARD_COLUMN_NAME + ");");
-                postHideDao.executeRawNoArgs("CREATE INDEX posthide_no_idx ON " + DatabaseHelper.POST_HIDE_TABLE_NAME + "(" + PostHide.NO_COLUMN_NAME + ");");
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_site_idx ON posthide (site);");
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_board_idx ON posthide (board);");
+                postHideDao.executeRawNoArgs("CREATE INDEX posthide_no_idx ON posthide (no);");
             } catch (SQLException e) {
                 Logger.e(TAG, "Error upgrading to version 27", e);
             }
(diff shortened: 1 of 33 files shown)


Adamantcheese:
This has LOTS of issues with crashing, especially when scrolling fast through a thread or reaching the last post in a thread (in which case it crashes immediately when a thread has only a few posts and can't scroll yet). As a note, you've removed a lot of the synchronized checks in the database methods which may be causing it.

K1rakishou:
@Adamantcheese Alright, I will look into it today.

K1rakishou:
When I tested this I had crashes in this place so I kinda closed it up with that check. Maybe it is related.
https://github.com/Floens/Clover/pull/654/files#diff-c43fe9f6e874c2b42ffaf3822bf843edR107

Adamantcheese:
It will also crash here with the opposite error as the one you fixed. https://github.com/K1rakishou/Clover/blob/(%23644)-recursive-filtering/Clover/app/src/main/java/org/floens/chan/ui/adapter/PostAdapter.java#L129
i.e.
```
if (!(holder instanceof StatusViewHolder)) {
                    // to avoid "PostViewHolder cannot be cast to StatusViewHolder" exception
                    return;
}
```

will also "solve" a problem, but the crashes will continue. It's likely that these two checks aren't needed and there's something else wrong with your code.

K1rakishou:
@Adamantcheese I think I figured out what is the problem.

https://github.com/K1rakishou/Clover/blob/(%23644)-recursive-filtering/Clover/app/src/main/java/org/floens/chan/ui/adapter/PostAdapter.java#L198

Here `displayList` is being accessed on the background thread and then there is the `notifyDataSetChanged` which will trigger `onBindViewHolder` to be called and it will access the `displayList` on the main thread. So, to fix this `displayList` should be synchronized. We can probably completely remove `sourceList` since it's not being used anywhere besides this place (and kinda pointless). The loop before `notifyDataSetChanged` call should be synchronized as well. This check (https://github.com/K1rakishou/Clover/blob/(%23644)-recursive-filtering/Clover/app/src/main/java/org/floens/chan/ui/adapter/PostAdapter.java#L107) is not needed anymore and should be removed as well. Other places that access `displayList` should be synchronized as well. I will push the fixes a little bit later today, have some stuff to do. Meanwhile you can just use this patch.

~~[#654_fix.zip](https://github.com/Floens/Clover/files/3077203/654_fix.zip)~~

Adamantcheese:
It's about 2AM for me, so whenever you get to it, I'll test it and make sure it's good before putting it in. 

Although why does this need to be done in a background thread? The current version works fine with it running on the main thread and the synchronization is going to be a pain in the ass to make work correctly (as I figured out when trying to resolve this).

If I remember right, most of the crashes were due to IndexOutOfBounds exceptions. It also looks like [another reason it might fail is here](https://github.com/K1rakishou/Clover/blob/(%23644)-recursive-filtering/Clover/app/src/main/java/org/floens/chan/ui/adapter/PostAdapter.java#L201) where the background thread is resetting the indicator position before calculating it, instead of using a thread-safe local variable to calculate it before setting it. That -1 being set for (relatively) quite a while, considering the holder stuff seems to wait in nanoseconds before failing, would explain those exceptions. If you do choose to continue using a background thread, you should probably double check that because it might cause crashes still.

K1rakishou:
@Adamantcheese 

>Although why does this need to be done in a background thread?

Before this task you could only hide threads. Hidden threads were stored in the database as well as in a cache variable for fast look-ups (https://github.com/Floens/Clover/pull/654/files#diff-d4c6b656f34af7234a6390b559ec3688L21). That's why it could work safely on the main thread. But it could only hold 250 threads at a time. After I've implemented the ability to hide posts as well as threads the number of 250 hidden posts became kinda small. So I've removed those caches and instead filtering is now reading hidden posts directly from the database. And now it can't run on the main thread since it will just crush (or just hung for a while if there is an ability to disable throwing an exception when accessing database from the main thread in the orm lite).

I think you are right that `lastSeenIndicatorPosition` may cause different crashes so what I think to do is to completely extract this code from the adapter (it shouldn't be there in the first place) to a different place and then just assign the filtering result into the adapter on the main thread. I'm gonna work on it later today.

K1rakishou:
Alright. This should fix it.

ghost:
> Hey @K1rakishou , some threads hide some don't.
> The message "Thread hidden" appears but the threads remains and is accesible.

DISREGARD WHAT I WROTE ABOVE.
I guess there was something wrong with the database because I deleted it and now everything is working as intended.

Also, could you make it that instead of removing the replies they would just hide?
Dashchan does it like you hide them but they stay in place and if you click them they appear again.

Dashchan-Hide

Astridchan:
Wouldn't it make sense to add options for "remove whole chain" and "hide whole chain" in the filter settings? As currently the recursive filtering feature is only available when manually hiding posts.

Screenshot_2019-04-15-12-39-42-755_org floens chan

Also, Muyli above has a point regarding hiding vs. removing posts. Currently the options in filter settings are "hide" and "remove", where "hide" will gray out the post and "remove" will completely remove it. But when you click the three button menu on a post, the "hide" option doesn't "hide" the post, it removes it. So the same term "hide" is being used for two different actions, which I think can lead to confusion. My suggestion would be to add both "remove" and "hide" in the post menu, and to keep the menu from getting cluttered, instead of having a "hide whole chain" and "remove whole chain" option separately, you would get a Y/N prompt "hide/remove whole chain?" when pressing to hide or remove a post with replies.

K1rakishou:
@myuli @Astridchan 
So, to sum it up. There have to be two menu items - "Hide" which will grey out a post and "Remove" which will completely remove a post like it is implemented right now. Upon clicking a post if it has replies to it a dialog window should pop up suggesting the ability to hide/remove the whole chain. Clicking on hidden (not removed) post should bring it back up. Is this correct?

>Wouldn't it make sense to add options for "remove whole chain" and "hide whole chain" in the filter 
settings? As currently the recursive filtering feature is only available when manually hiding posts.

Wouldn't people start complaining that posts are disappearing without any reason because they forgot this option is turned on?

Astridchan:
@K1rakishou 
> So, to sum it up. There have to be two menu items - "Hide" which will grey out a post and "Remove" which will completely remove a post like it is implemented right now. Upon clicking a post if it has replies to it a dialog window should pop up suggesting the ability to hide/remove the whole chain. Clicking on hidden (not removed) post should bring it back up. Is this correct?

Exactly.

> Wouldn't people start complaining that posts are disappearing without any reason because they forgot this option is turned on?

Well that's on them, isn't it?

I'll try to explain why I think this feature would be useful, and why I would want to see it added. Basically, in a few of the generals I frequent, there are these copy-pasta shitposts that get posted in every new thread for (You)s. I've got those posts filtered so that they're automatically removed, but the replies remain, so I still end up seeing the shitposts when I follow back the reply chains. A "remove whole chain" would make it so both the shitpost and the replies to it get removed.

EDIT:
One thing I forgot to mention is that currently, if I choose "remove whole chain" on a post, new replies to that post will still show up. To me it would make sense that any subsequent reply to a post that has been removed/hidden should also be removed/hidden.

EDIT2: I'll give an example for this too. Say someone posts one of those "Roll for X" images that tend to get a lot of "roll" replies. Currently, if you chose "remove whole chain" you'd only remove the post and the replies made *up to that point*, but all future "roll" posts will still show up. What you'd want to happen is that all current *and* future replies to the post get removed.

K1rakishou:
The last commit is Work In Progress but you can already test it if you would like (I would really appreciate it). But do not merge it yet.

It "should" work. 
For now all posts that have replies to a post that was matched by a filter will be hidden (greyed out). I'm gonna need to store the filter id along with the posts to figure out what to do with such a post (hide/remove or highlight) when filtering them.

K1rakishou:
Alright, looks good.
So, now there are two menu options - hide and remove. 
- If a post has replies to it then additional dialog will be shown with suggestion to remove/hide the whole chain. 
- If you choose to remove a single post and then someone replies to this post the reply will NOT be hidden (idk should I implement it?).
- If a post has replies and you choose to hide/remove the whole chain then all new replies to any of the posts from the chain will be hidden/removed as well.
- Filters now have an additional parameter "Apply to replies". 

- If anyone replies to a post that is hidden by a filter with this flag it will be hidden/removed/highlighted as well. The action (hide/remove/highlight) is inherited from the parent post. So if you have a filter that highlights posts and someone replies to a post that was hidden by that filter that post will be highlighted as well.
- Hide/remove menu options are disabled for posts that has been affected by a filter (because it is hard to figure out what to do with reply chains when one of the posts was hidden or removed by the menu option).

I changed the DB again and have no idea how to handle exporting/importing from #678 so that @Adamantcheese build works because I will need to merge this task with #678 but then this task will grow up in size. We really need @Floens to start merging PRs.

Clover/app/src/main/java/org/floens/chan/core/site/parser/PostParseCallable.java
```diff
@@ -77,15 +77,18 @@ private void processPostFilter(Post.Builder post) {
             Filter filter = filters.get(i);
             if (filterEngine.matches(filter, post)) {
                 FilterEngine.FilterAction action = FilterEngine.FilterAction.forId(filter.action);
+
+                // TODO: store filter id in the post so we can figure out what exactly we want to do
```
K1rakishou: should remove this

Adamantcheese:
I'll take a look at what changed in the db and do what I need to for import/export. I have an exported file for the last version that I can test with.

Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHideManager.java
```diff
@@ -1,118 +1,399 @@
...
+
+                helper.postHideDao.createIfNotExists(postHide);
             }
 
             return null;
         };
     }
 
-    public Callable<Void> clearAllThreadHides() {
+    public Callable<Void> removeThreadHide(PostHide hide) {
```
K1rakishou: Probably should do the same as in removePostsHide

Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHideManager.java
```diff
@@ -1,118 +1,399 @@
...
+                postWithAllReplies.add(newPost);
+            }
+        }
+    }
+
+
+    /**
+     * Rebuilds a child post with custom filter parameters
+     * */
+    private Post rebuildPostWithInheritedFilter(
```
K1rakishou: rename to rebuildPostWithCustomFilter

Clover/app/src/main/java/org/floens/chan/core/model/Post.java
```diff
@@ -244,6 +250,34 @@ public PostImage image() {
...
+        return 31 * no;
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (other == null) {
+            return false;
+        }
+
+        if (this.getClass() != other.getClass()) {
```
K1rakishou, line 275: Add this == other check

Clover/app/src/main/java/org/floens/chan/core/model/Post.java
```diff
@@ -244,6 +250,34 @@ public PostImage image() {
...
     }
 
+    @MainThread
+    public boolean hasFilterParameters() {
+        return filterRemove || filterHighlightedColor != 0 || filterReplies || filterStub;
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * no;
```
K1rakishou: Add board and site as well

Clover/app/src/main/java/org/floens/chan/core/model/Post.java
```diff
@@ -244,6 +250,34 @@ public PostImage image() {
...
+    public boolean equals(Object other) {
+        if (other == null) {
+            return false;
+        }
+
+        if (this.getClass() != other.getClass()) {
+            return false;
+        }
+
+        return this.no == ((Post) other).no;
```
K1rakishou: Add board and site as well

Clover/app/src/main/java/org/floens/chan/core/model/Post.java
```diff
@@ -244,6 +250,34 @@ public PostImage image() {
...
+    }
+
+    @Override
+    public String toString() {
+        return "[no = " + no + ", comment = " + comment + "]";
```
K1rakishou: Add board and site as well

Clover/app/src/main/java/org/floens/chan/utils/PostUtils.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class PostUtils {
+
+    private PostUtils() {
+    }
+
+    public static Post findPostById(int id, ChanThread thread) {
```
K1rakishou, line 21: Maybe pass a list of posts here instead of ChanThread so we can get rid of ChanThread dependency here?

Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java
```diff
@@ -256,6 +259,36 @@ public void onUpgrade(SQLiteDatabase database, ConnectionSource connectionSource
                 Logger.e(TAG, "Error upgrading to version 25", e);
             }
         }
+
+        if (oldVersion < 26) {
+            try {
+                postHideDao.executeRawNoArgs("ALTER TABLE threadhide RENAME TO " + POST_HIDE_TABLE_NAME + ";");
```
K1rakishou: Woah, this is actually bad. Database migration queries should not have variables in them. If anyone renames at least one of them everything will break. Should replace them all. Idk how I didn't notice it before.

Adamantcheese:
@K1rakishou looks like from the last build of mine to the most recent one, no import/export changes needed to be made so the import/export version stayed the same. I've also already replaced all the database variable names in my build with their respective strings.

K1rakishou:
@Adamantcheese I see you have removed them from everywhere. I actually meant to just remove them from this class (https://github.com/Adamantcheese/Clover/blob/92a7e6d05778c6f304b4118be9c9de40d519a9e7/Clover/app/src/main/java/org/floens/chan/core/database/DatabaseHelper.java) because they should never change in migrations to avoid crashes, but keep them everywhere else to avoid mistakes when typing them by hand every time. And it is really helpful when you want to rename some column because you will only have to rename it in one place.

Adamantcheese:
I figured if everything was hardcoded it would be better and these names shouldn't really ever change.

K1rakishou:
@Adamantcheese Well normally yes, they should not change. Well I guess it's fine.

K1rakishou:
Regarding the importing/exporting I guess it is fine too because we have just merged the importing/exporting so there is just nothing to import from.

Adamantcheese:
The import/export was merged before this last set of changes you made, and the GSON class maker thing defaults the field to false if it didn't exist, so there's no upgrade needed.

Adamantcheese (changes requested):
Filter has an applyToReplies variable that is annotated as a database field, but the database column name specified in DatabaseHelper is apply_to_replies. If the actual variable name doesn't match the column name, you MUST specify the column name in the @DatabaseField annotation, otherwise the database engine will take the literal variable name as the column name. [See this commit for the fix.](https://github.com/Adamantcheese/Clover/commit/bfeed0075c8bfe2d287f585694e881fcf49e59f3)

Not only did this crash my build several times with weird, confusing errors, but also you'd never even be reading from the right table!

Clover/app/src/main/java/org/floens/chan/core/model/orm/Filter.java
```diff
@@ -49,6 +50,12 @@
     @DatabaseField(canBeNull = false)
     public int color;
 
+    @DatabaseField(canBeNull = false)
```
Adamantcheese: I meant to point to this line of code, sorry.

Adamantcheese (changes requested):

K1rakishou:
@Adamantcheese Damn, I guess I forgot to specify the name in the annotation. I wonder why it didn't crash for me when I was testing it.

Clover/app/src/main/java/org/floens/chan/utils/PostUtils.java
K1rakishou: Nah, it's more convenient this way.

Clover/app/src/main/java/org/floens/chan/core/model/orm/Filter.java
K1rakishou: Done.

ghost:
Yo @K1rakishou can you make it that hidden threads or replies unhide when you click on them?

K1rakishou:
@myuli Sure, but only for the posts that you hid by yourself and not by the filters. That's because of how filters work right now. I guess it should also be implemented for removed threads/posts.

K1rakishou:
Alright. You can now unhide threads/posts by clicking on them. 
There is also a new menu item that will show all posts that were removed in a thread.

ghost:
@K1rakishou Thanks, man. 
@Adamantcheese Can you add this?

Issue #468, opened 2018-04-26
Request: confirmation when clearing all pinned threads


bakugo:
Twice now I have deleted hundreds of pinned threads by accidentally pressing the clear all button instead of the watch toggle button.

I'm aware that an "undo" button pops up for a few seconds, but when clearing many pins, the app will freeze for a significant amount of time and a "Clover is not responding" overlay will often pop up. If you're not ready to press "undo" as soon as it unfreezes, the pins are gone and there's no way to get them back.

Floens:
I think it's better to not let Clover hang on deleting so many pins.

PR #655, opened 2019-03-30
#468, Speed up pins deleting

K1rakishou:
Relates to #468. 
Should improve the speed of deleting hundreds of pins at a time.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
@@ -224,6 +224,23 @@ public class WatchManager {
         EventBus.getDefault().post(new PinRemovedMessage(pin));
     }
 
+    public void deletePins(List<Pin> pinList) {
+        for (Pin pin : pinList) {
+            pins.remove(pin);
+            destroyPinWatcher(pin);
+        }
+
+        databaseManager.runTask(databasePinManager.deletePins(pinList));
+
+        // Update the new orders
+        sortListAndApplyOrders();
+        updatePinsInDatabase();
+
+        updateState();
+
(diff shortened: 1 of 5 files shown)


Clover/app/src/main/java/org/floens/chan/core/manager/WatchManager.java
```diff
@@ -224,6 +224,23 @@ public void deletePin(Pin pin) {
...
+
+        databaseManager.runTask(databasePinManager.deletePins(pinList));
+
+        // Update the new orders
+        sortListAndApplyOrders();
+        updatePinsInDatabase();
+
+        updateState();
+
+        EventBus.getDefault().post(new AllPinsRemovedMessage());
```
Adamantcheese, line 241: This event bus message is not correct unless pinList is equivalent to WatchManager's pins variable. This will cause the error noted here: https://github.com/Adamantcheese/Clover/issues/2.

Clover/app/src/main/java/org/floens/chan/ui/adapter/DrawerAdapter.java
```diff
@@ -223,6 +223,11 @@ public void onPinRemoved(Pin pin) {
         notifyDataSetChanged();
     }
 
+    public void onAllPinsRemoved() {
```
Adamantcheese, line 226: Perhaps use the onPinsChanged message instead of this one.

Adamantcheese (changes requested):
In response to https://github.com/Adamantcheese/Clover/issues/2.

K1rakishou:
I tried to use your fix but apparently it depends on the other task that is merged in your branch and I don't have it. So I guess it will be just easier to use yours.

Issue #375, opened 2017-12-24
Ability to swipe an image


OpenUp596:
You can close an image by clicking on it but some people(me included) are used to swipe an image(vertically) to close it.

Not saying this is a better wat but as an option in settings it would be great.

Thank you very much for the great app, btw.

droidberg:
actually, this feature could be extended. maybe swipe up to close and swipe down to save? its far easier and more intuitive imho.

ridhozhr10:
well maybe theres a tab container on the bottom screen, so when you swipe down its just hide the previous image, and when you swipe up the tab you open it again. i think its useful

Adamantcheese:
I gave it a shot, but the current ViewPager implementation eats onDown motion events before I can figure out if a swipe up/down has occurred. 
[A possible solution is to use a RecyclerView with a SnapHelper, but that requires API 25 and isn't in API 15. I'd like to avoid having to upgrade to a higher API level.](https://proandroiddev.com/looks-like-a-viewpager-sounds-like-a-viewpager-its-a-recyclerview-505cdf129b58) 

[Another possible solution is to use a multidirectional view pager, but I don't know if that would cause issues; it would also be adding in deprecated code.](https://github.com/JakeWharton/Android-DirectionalViewPager)

PR #656, opened 2019-03-31
Add rotate options to image view

Adamantcheese:
Like with #375 a better implementation will probably work better, as this should be a double touch rotate option instead of a menu option


Astridchan:
A small suggestion: images that you rotate should keep the orientation they're rotated to. Currently if you close the image, or swipe to another image, the orientation will reset.

Adamantcheese:
That would be more effort than what I'm willing to put in for a feature like this, and as I say above, a double touch rotate would be better.

PR #665, opened 2019-04-06
Add Wired-7 support

Wired-7:
Wired-7 is a vichan-lainchan imageboard.
Posting working.


qorg11:
node se te ha olvidado darnos la apk

Issue #677, opened 2019-04-14
[BUG] App crashes on any android below Marshmallow because of a unicode symbol


K1rakishou:
At least on the emulator. Here is the stacktrace:

```
04-14 16:05:13.229 4282-4282/? A/art: art/runtime/check_jni.cc:65] JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal start byte 0xf0
04-14 16:05:13.229 4282-4282/? A/art: art/runtime/check_jni.cc:65]     string: '😴'
04-14 16:05:13.229 4282-4282/? A/art: art/runtime/check_jni.cc:65]     in call to NewStringUTF
04-14 16:05:13.229 4282-4282/? A/art: art/runtime/check_jni.cc:65]     from java.lang.String android.content.res.StringBlock.nativeGetString(long, int)
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65] "main" prio=5 tid=1 Runnable
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   | group="main" sCount=0 dsCount=0 obj=0x7492dfa8 self=0x7f5195197000
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   | sysTid=4282 nice=0 cgrp=apps sched=0/0 handle=0x7f519852c6a0
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   | state=R schedstat=( 0 0 0 ) utm=76 stm=7 core=0 HZ=100
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   | stack=0x7fff814c3000-0x7fff814c5000 stackSize=8MB
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   | held mutexes= "mutator lock"(shared held)
04-14 16:05:13.230 4282-4282/? A/art: art/runtime/check_jni.cc:65]   native: #00 pc 00006c19  /system/lib64/libbacktrace_libc++.so (UnwindCurrent::Unwind(unsigned long, ucontext*)+57)
[...]

Issue #674, opened 2019-04-13
[BUG] Pinned thread title is not saved to the database when pinning a thread from a catalog


K1rakishou:
![234](https://user-images.githubusercontent.com/10332458/56080367-0601ab00-5e09-11e9-93c0-f8b4cf3690fb.png)

This happens because when pinning a thread we haven't visited before (no loadable for that thread stored in the database) we don't assign the thread title for the loadable associated with it. 

https://github.com/Floens/Clover/blob/4b57ad38988c01aa764ae139c73b981ab7803d14/Clover/app/src/main/java/org/floens/chan/core/presenter/ThreadPresenter.java#L540

How to fix:
Assign a title to the loadable before calling `databaseManager.getDatabaseLoadableManager().get()`
```
String title = PostHelper.getTitle(post, loadable);
Loadable pinLoadable = databaseManager.getDatabaseLoadableManager().get(
        Loadable.forThread(loadable.site, post.board, post.no, title)
);
watchManager.createPin(pinLoadable, post);
```

K1rakishou:
I'll probably fix it myself along with another task unless somebody does it faster than me.

Issue #289, opened 2017-02-11
Request: export pinned threads


wrsg:
Please create a feature to export (and maybe) import pinned threads, to make it easier to check on 404 threads from an archive later on

yellowjelly:
I desperately need this tbh.

bahari15:
bumping this thread, because I also really want to be able to export and import both pinned and bookmarked threads so it could be used as my backup data

PR #678, opened 2019-04-14
#638, Settings/pins/hidden posts/filters etc. import/export

### LINKED ISSUE #638: Thread and settings backup and restore function
https://github.com/Floens/Clover/pull/636#issuecomment-469944155

One feature that would be great to see is a thread and setting backup and restore function, like how the Reddit [Slide app](https://github.com/ccrama/Slide) does it (I believe [this](https://github.com/ccrama/Slide/blob/master/app/src/noGPlay/java/me/ccrama/redditslide/Activities/SettingsBackup.java) and [this](https://github.com/ccrama/Slide/blob/master/app/src/main/java/me/ccrama/redditslide/SettingValues.java) are the code files for this function). The data is stored in an xml file. Clover could easily have the same function with a little bit of code :)

K1rakishou:
Closes #638, #289 and couple of bugs #674 and #677.
This is a big boy and needs good testing.


Adamantcheese:
Couple of things:
1) [It appears there's already a restart function call in StartActivity, is there any particular reason you need to have a new one that looks very similar to this one? It's used for when the Theme changes.](https://github.com/Floens/Clover/blob/c8cdf29d030bb37922a9d8492c667e60c6ac08de/Clover/app/src/main/java/org/floens/chan/ui/activity/StartActivity.java#L306)
2) When opening up the import/export settings dialog for the first time, it doesn't prompt for getting the permission. This causes it to fail to make the proper folder and even if you grant access in the next step to export, it'll fail to export. If you exit out of the menu and go back into import/export again, it'll work correctly.
3) Also is there any particular reason to remove all the `@Singleton` annotations? I think that may cause some serious issues by itself.
4) You've added this huge block in ChanSettings basically re-enumerating all the settings already listed. Maybe using https://developer.android.com/reference/android/content/SharedPreferences.html#getAll() would be a better method?

K1rakishou:
@Adamantcheese 
1. IIRC it didn't work for me when I tried it but maybe it was because I did something wrong. Alright, I will try that method again and if it works I will remove the one I've added.
2. Yes, this is a problem and I'll fix it.
3. I didn't just remove them, actually I moved the dependency initialization here `Clover/app/src/main/java/org/floens/chan/core/di/` (https://github.com/Floens/Clover/pull/678/files#diff-d6a720136bce5da9ab5e53dc28f9577cR137)
I was getting some strange exception (Feather cannot inject databaseManager to the Chan class) so I moved dependencies around until they all ended up in one place (The exception was actually because I had removed the default constructor from the SiteModel but Feather absorbs the stacktrace so it wasn't obvious at all). I can revert that change, but I personally think it is better when all dependencies are being initialized in Modules.
4. Yes, I did forget that you can just read them all with getAll(). I'll change it.

K1rakishou:
@Adamantcheese 
1. The other restart method didn't work for me for some reason.
2. Fixed.
3. 
4. Looks like it is impossible to deserialize shared prefs from a string without enumerating the whole collection of settings because you have to know at least what type the value is. I mean it's pretty easy to serialize them to a string (just check what instance the value is) but is kinda hard to deserialize. Primarily because some settings contain only numbers but their type is StringSetting (like this one: `public static final StringSetting fontSize`). So I have to leave it like it is now (with huge  list of all possible settings).

K1rakishou:
Oh, wait a  minute. It looks like I can just read the whole file directly without using the SharedPrefs class and then just write it back to the file the same way. I'm gonna experiment with this tomorrow.

Adamantcheese:
Cool, thanks for all the updates. I'll wait another little bit to merge it in, but everything looks fine. I'll test it properly when you do that last change for 4.

K1rakishou:
@Adamantcheese 
Alright, done. 
Now there is one thing I need to mention: when adding/deleting columns in the database do not forget to handle that in ImportExportRepository onUpgrade/onDowngrade methods. Otherwise it may broke something for the user. The settings file also has it's own version (`private int version = CURRENT_EXPORT_SETTINGS_VERSION`). So don't forget to change it as well when changing something in the export schema. Sometimes it will not be possible to restore some settings from the file when upgrading/downgrading. Like when I added "whole_thread" column to the ThreadHide table. It is impossible to figure out this parameter on it's own (and it is impossible to use some default value as well) so the only thing that can be done here is to completely clear all the ThreadHides. I have no idea how to mitigate this issue. The only thing is probably to not change the database columns and when you do - notify the user about it.

Adamantcheese:
Adding a comment at the end of DatabaseHelper's onUpgrade method to say that ImportExportRepository's onUpgrade method should be modified if needed if onUpgrade changes should be enough warning for other devs.
I don't believe the onDowngrade method will ever be used though. The database only has an onUpgrade method because the database version should only ever go up. The same should apply to the export settings version. It should probably be removed to avoid confusion.
And now that I'm looking at it, you parse the file with the fromJson(String, Class) method. If the exported class changes (or any nested classes) and you try to use that, I think it'll throw an exception and not allow an import? It should probably log "Error while trying to import pins", which isn't even a correct error message for what either the import or export functions do anymore.

K1rakishou:
@Adamantcheese 
>Adding a comment at the end of DatabaseHelper's onUpgrade method to say that ImportExportRepository's onUpgrade method should be modified if needed if onUpgrade changes should be enough warning for other devs.

I agree. Will add it.

>I don't believe the onDowngrade method will ever be used though. The database only has an onUpgrade method because the database version should only ever go up. The same should apply to the export settings version. It should probably be removed to avoid confusion.

Yeah, probably it won't be used. So in case of downgrade we can just notify the user that nothing will be imported and do nothing. But I don't think it is a good idea to remove the version. This way users won't be able to migrate their settings from one version to another when something in the database changes (or it may be even worse, like, the settings will get applied without exceptions but they will fuck up something else and user won't be able to use the app). 

>And now that I'm looking at it, you parse the file with the fromJson(String, Class) method. If the exported class changes (or any nested classes) and you try to use that, I think it'll throw an exception and not allow an import? 

No it will just use a default value (like 0 for ints or null for strings or any other object).

And that's why we need the version and onUpgrade method where you can do something with data that couldn't be imported due to changes.

Adamantcheese:
I've taken a good amount of time to look at it, and I can't see any issues anymore. I've merged it in with my fork.

K1rakishou:
@Adamantcheese Thanks.

Clover/app/src/main/java/org/floens/chan/ui/controller/export/ImportExportSettingsController.java
```diff
@@ -0,0 +1,306 @@
...
+    private void onExportClicked() {
+        String state = Environment.getExternalStorageState();
+        if (!Environment.MEDIA_MOUNTED.equals(state)) {
+            showMessage(context.getString(R.string.error_external_storage_is_not_mounted));
+            return;
+        }
+
+        getPermissionHelper().requestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE, granted -> {
+            if (granted) {
+                navigationController.presentController(loadingViewController);
```
K1rakishou: Probably should move this inside the check below because if the presenter is not set then we will show the loading view but won't start the exporing so it will just be shown forever.

Clover/app/src/main/java/org/floens/chan/ui/controller/export/ImportExportSettingsController.java
```diff
@@ -0,0 +1,306 @@
...
+                .setMessage(warningMessage)
+                .setPositiveButton(R.string.continue_text, (dialog1, which) -> onStartImportButtonClicked())
+                .setNegativeButton(R.string.cancel, null)
+                .create();
+
+        dialog.show();
+    }
+
+    private void onStartImportButtonClicked() {
+        navigationController.presentController(loadingViewController);
```
K1rakishou: Same as above. This needs to be moved inside the check below.

Clover/app/src/main/java/org/floens/chan/ui/controller/export/LoadingViewController.java
```diff
@@ -0,0 +1,50 @@
...
+import android.app.Activity;
+import android.content.Context;
+import android.os.Build;
+import android.view.Window;
+
+import org.floens.chan.R;
+import org.floens.chan.controller.Controller;
+import org.floens.chan.utils.AndroidUtils;
+
+public class LoadingViewController extends Controller {
```
K1rakishou: Like 99% of this class code can be extracted to an abstract class and then other controllers that use dark transparent background can inherit from it.

Clover/app/src/main/res/values/strings.xml
```diff
@@ -577,4 +589,20 @@ Don't have a 4chan Pass?<br>
...
+    <string name="import_or_export_settings" tools:ignore="MissingTranslation">Import/Export settings</string>
+    <string name="export_settings" tools:ignore="MissingTranslation">Export settings</string>
+    <string name="export_settings_to_a_file" tools:ignore="MissingTranslation">Export settings to a file</string>
+    <string name="import_settings" tools:ignore="MissingTranslation">Import settings</string>
+    <string name="import_settings_from_a_file" tools:ignore="MissingTranslation">Import settings from a file</string>
+    <string name="import_warning_title" tools:ignore="MissingTranslation">Warning</string>
+    <string name="import_warning_text" tools:ignore="MissingTranslation">You are about to import settings/saved pins/filters/hidden posts from a file. The file must be located at \"%1$s\" and have a name \"%2$s\". This will clear all your existing settings/pins/threads and replace them with the ones from the file. You MAY LOOSE some of your settings/pins/hidden threads if your are trying to import a file from with different app version (Upgrade or downgrade). The app will be restarted. \nAre you sure you want to continue?</string>
```
K1rakishou: Correct this message.

Clover/app/src/main/java/org/floens/chan/core/repository/ImportExportRepository.java
```diff
@@ -0,0 +1,476 @@
...
+                }
+
+                if (!settingsFile.exists() || !settingsFile.canWrite()) {
+                    throw new IOException(
+                            "Something wrong with export file (Can't write or it doesn't exist) "
+                                    + settingsFile.getAbsolutePath()
+                    );
+                }
+
+                try (RandomAccessFile raf = new RandomAccessFile(settingsFile, "rw")) {
```
Adamantcheese, line 127: This doesn't work correctly with UTF-8 encoded strings. Some symbols tend to get fucked up and then you won't be able to import afterwards. Use a FileWriter instead, which writes out UTF-8 by default and doesn't cause issues later on. Noticeably, ™ seems to turn into " for some reason?

Adamantcheese (changes requested):

Clover/app/src/main/java/org/floens/chan/core/repository/ImportExportRepository.java
K1rakishou: Yeah, apparently for UTF to work it should be writeUtf() not writeBytes().

K1rakishou:
Should be fixed now. Tested on bookmarked threads with japanese characters in the title.
Btw, I have changed the CURRENT_EXPORT_SETTINGS_VERSION to version 2 and disabled upgrading from v1 to v2 because it will take a lot of additional code to deserialize v1 settings on v2. So when releasing a new version of the app we probably should notify the users (in the changelist) that they should re-export their settings again.

Clover/app/src/main/java/org/floens/chan/core/presenter/ImportExportSettingsPresenter.java
```diff
@@ -40,80 +42,89 @@
...
 
             @Override
-            public void onError(Throwable error, ImportExportRepository.ImportExport importExport) {
+            public void onExportError(Throwable error) {
                 //called on background thread
 
-                if (callbacks != null) {
-                    callbacks.onError("Error while trying to export settings = " + error.getMessage());
+                if (exportCallbacks != null) {
+                    exportCallbacks.onError("Error while try to export settings: " + error.getMessage());
```
K1rakishou: "trying".

Clover/app/src/main/java/org/floens/chan/core/repository/ImportExportRepository.java
```diff
@@ -60,7 +59,7 @@
 
     // Don't forget to change this when changing any of the Export models.
     // Also, don't forget to handle the change in the onUpgrade or onDowngrade methods
```
K1rakishou: Remove "or onDowngrade" since it is not supported.

Clover/app/src/main/java/org/floens/chan/core/repository/ImportExportRepository.java
```diff
@@ -295,7 +294,20 @@ private void writeSettingsToDatabase(@NonNull ExportedAppSettings appSettingsPar
...
-    private ExportedAppSettings onUpgrade(int version, ExportedAppSettings appSettings) {
+    @NonNull
+    private ExportedAppSettings onUpgrade(
+            int version, ExportedAppSettings appSettings) throws UpgradeImpossibleException {
+        if (version < 2) {
+            ///////////////////////////////////////
+            // Upgrade from version 1 to version 2.
+            // In version 2 we had to change the way the settings are serialized to a file because
+            // it was not considering the UTF-8 characters. This breaks everything. So we have no other
+            // way other than clear everything.
```
K1rakishou: "other than not import anything"

Adamantcheese: This is entirely unnecessary because the exception being thrown will be in line 151. The actual JSON fails to parse because of mismatched quotes, so you'd never even get to this point. There's no need to increment the version number.

Adamantcheese (changes requested):

K1rakishou:
Yeah, that's a problem. I think I will have to rewrite the structure of the ExportedSettings so that the version field is not a part of the json. Because right now we need to read the json just to get a version. And because of that we can't check the file version before we start parsing. I don't want to leave it as it is because the error message in this case is not really clear to the user.

K1rakishou:
Alright. I have moved out the settings file version from the json to the file header so we can now check the version before trying to parse the json (in case it breaks again). Also, when settings file version changes a suggestion to re-export settings file will be shown to the user so they can keep their settings as updated as possible (probably nobody will do that, but just in case).

Clover/app/src/main/java/org/floens/chan/ui/controller/export/ImportExportSettingsController.java
```diff
@@ -0,0 +1,322 @@
...
+            onPermissionGrantedForDirectoryCreation();
+            return;
+        }
+
+        // Ask the user's permission to check whether the Clover directory exists and create it if it doesn't
+        AlertDialog dialog = new AlertDialog.Builder(context)
+                .setTitle(context.getString(R.string.clover_directory_may_not_exist_title))
+                .setMessage(context.getString(R.string.clover_directory_may_not_exist_message))
+                .setPositiveButton(context.getString(R.string.create), (dialog1, which) -> {
+                    getPermissionHelper().requestPermission(Manifest.permission.READ_EXTERNAL_STORAGE, granted -> {
```
Adamantcheese: This should be WRITE_EXTERNAL_STORAGE, because we're making the directories here which is writing, not reading.

K1rakishou: Huh, that's funny, because couple of lines above I'm checking whether we already have WRITE_EXTERNAL_STORAGE and if not then for some reason requesting the READ_EXTERNAL_STORAGE.

PR #697, opened 2019-04-28
#696, New captcha window

### LINKED ISSUE #696: [FEATURE] Implement new captcha window as a replacement for the webview one
- Should be a custom view that can resize to fit the phone screen. 
- The whole image should be clickable instead of how it is now in the webview captcha where you can only click checkboxes.
- Should use no-js captcha (well there is no other way anyway)
- Should fallback to webview captcha when some error has occurred while trying to parse google captcha data.

K1rakishou:
Closes #696 
For some reason it takes me like 20 tries to solve the captcha maybe because I tested it a lot today and it just doesn't like me now. Or maybe I forgot to add something (I did add the SID, HSID, SSID and NID cookies though). It should automatically fallback to the webview captcha if something couldn't be parsed (or you can use the button).


Clover/app/src/main/java/org/floens/chan/ui/captcha/v2/CaptchaNoJsHtmlParser.java
```diff
@@ -0,0 +1,398 @@
...
+                            row * imageHeight,
+                            imageWidth,
+                            imageHeight);
+
+                    resultImages.add(imagePiece);
+                }
+            }
+
+            return resultImages;
+        } finally {
```
K1rakishou, line 395: Should recycle bitmaps (in the catch clause) that were already added to the list when an exception occurs in the process of decoding the images. Right now they will leak.

Clover/app/src/main/java/org/floens/chan/ui/captcha/v2/CaptchaNoJsLayoutV2.java
```diff
@@ -0,0 +1,274 @@
...
+                        R.string.captcha_layout_v2_captcha_request_is_already_in_progress));
+                break;
+        }
+    }
+
+    @Override
+    public void onCaptchaInfoParsed(CaptchaInfo captchaInfo) {
+        // called on a background thread
+
+        AndroidUtils.runOnUiThread(() -> {
```
K1rakishou, line 170: Should probably disable the Verify button until we have the CaptchaInfo. Do not forget to reset the button's disabled state in initialize/onError/onVerificationDone/onDestroy.

Clover/app/src/main/java/org/floens/chan/ui/captcha/v2/CaptchaNoJsPresenterV2.java
```diff
@@ -0,0 +1,347 @@
...
+                    .header("Accept-Language", "en-US")
+                    .header("Cookie", googleCookies)
+                    .build();
+
+            okHttpClient.newCall(request).enqueue(new Callback() {
+                @Override
+                public void onFailure(Call call, IOException e) {
+                    if (callbacks != null) {
+                        try {
+                            callbacks.onCaptchaInfoParseError(e);
```
K1rakishou: Set prevCaptchaInfo to null here

Clover/app/src/main/java/org/floens/chan/ui/captcha/v2/CaptchaNoJsPresenterV2.java
```diff
@@ -0,0 +1,347 @@
...
```
K1rakishou: Set prevCaptchaInfo to null here

Clover/app/src/main/java/org/floens/chan/ui/captcha/v2/CaptchaNoJsPresenterV2.java
```diff
@@ -0,0 +1,347 @@
...
+                return null;
+            } else {
+                // got the challenge
+                CaptchaInfo captchaInfo = parser.parseHtml(bodyString, siteKey);
+
+                if (callbacks != null) {
+                    callbacks.onCaptchaInfoParsed(captchaInfo);
+                }
+
+                return captchaInfo;
```
K1rakishou, line 458: Probably should return null here to reset prevCaptchaInfo when callbacks are null. Otherwise we are kinda stuck in a state where there is no captcha window (But I guess reset should fix it?).

Clover/app/src/main/java/org/floens/chan/ui/layout/ReplyLayout.java
```diff
@@ -269,7 +272,25 @@ public void initializeAuthentication(Site site, SiteAuthentication authenticatio
                     break;
                 }
                 case CAPTCHA2_NOJS:
-                    authenticationLayout = new CaptchaNojsLayout(getContext());
+                    if (!useV2NoJsCaptcha) {
```
K1rakishou: Probably should invert the condition here

Clover/app/src/main/java/org/floens/chan/ui/layout/ReplyLayout.java
```diff
@@ -269,7 +272,25 @@ public void initializeAuthentication(Site site, SiteAuthentication authenticatio
...
+                        // default webview-based captcha view
+                        authenticationLayout = new CaptchaNojsLayoutV1(getContext());
+                    } else {
+                        // new captcha window without webview
+                        authenticationLayout = new CaptchaNoJsLayoutV2(getContext());
+                    }
+
+                    ImageView resetButton = captchaContainer.findViewById(R.id.reset);
+                    if (resetButton != null) {
+                        if (!useV2NoJsCaptcha) {
```
K1rakishou: Probably should invert the condition here

Clover/app/src/main/res/layout/layout_captcha_challenge_image.xml
```diff
@@ -0,0 +1,29 @@
...
+    <android.support.v7.widget.AppCompatImageView
+        android:id="@+id/captcha_challenge_blue_checkmark"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
+        android:layout_marginStart="8dp"
+        android:layout_marginLeft="8dp"
+        android:layout_marginTop="8dp"
+        android:visibility="gone"
+        android:src="@drawable/ic_blue_checkmark_24dp"
+        app:layout_constraintStart_toStartOf="@+id/captcha_challenge_image"
```
K1rakishou: Add #FFF tint here

K1rakishou: Implemented it the other way

Adamantcheese:
@K1rakishou I submitted a PR to fix the checkmark not being centered, and having to use an extra background view to make it white due to transparency.

K1rakishou:
@Adamantcheese 
>checkmark not being centered

But that's how it is implemented in the google's recaptcha window. You think it should be centered?



>to make it white due to transparency

And this one already fixed here - https://github.com/Floens/Clover/pull/697/commits/2820d23ff787e7d97cb69c0536ea5e2d08c8e3fb#diff-460f4eca5fc92d000ad076afc5756ccaR31

Btw, do you have to enter the captcha like 10-15 times until it is accepted too? I even changed the cookies to the one dashchan uses but it still forces me to enter the captcha a lot of times. Apparently something is missing, but I can't figure out what exactly.

Adamantcheese:
The white thing I changed so it's part of the drawable and doesn't use an underlying view to make it white, because that's just silly. And for me that check mark was appearing... elsewhere? Here's a photo of the Pixel 2 emulator from when I tested it. It looks like you've crafted it specifically for the second style of captcha and not the first one, and I pretty much only ever get the 3x3 kind and not the 4x4 kind.

It's easier to center it anyways, I don't super care about looks for it.

Adamantcheese:
And yeah, I had to enter the Captcha like 5-6 times.

K1rakishou:
@Adamantcheese 
>The white thing I changed so it's part of the drawable and doesn't use an underlying view to make it white, because that's just silly.

Oh yeah, this makes sence.

>And for me that check mark was appearing... elsewhere? Here's a photo of the Pixel 2 emulator from when I tested it. It looks like you've crafted it specifically for the second style of captcha and not the first one, and I pretty much only ever get the 3x3 kind and not the 4x4 kind.

Well that's strange because I only get the 3x3 captchas too and the checkmark is where it is supposed to be (upper left corner).

K1rakishou:
Alright then I will merge it later today. Still we have to do something with the amount of captcha attempts it takes to make a post.

Adamantcheese:
It could be that, if we're both testing on emulators, it's just giving us more to ensure that we're not robots. If you're testing on an actual device though, yeah, it's a problem.

K1rakishou:
@Adamantcheese 
Over the last week I tried it on the real device too and most of the time it gives you the same amount of challenges. But sometimes it just accepts the first one.

Adamantcheese:
Weird. I don't know anything about how Captcha works though, so I won't be much help.

K1rakishou:
Well, I guess for now we can just make it opt-out.
The other idea I have is to get the real google cookies upon the first app run, then store them somewhere and use them instead of the hardcoded ones.

Adamantcheese:
From a little research it looks like the cookies might be per-user, which would make the hardcoded ones a problem anyways because anyone using the app would be using the same cookie and they might be around the world, which looks sketchy as hell.

K1rakishou:
Yeah, probably. But on the other hand dashchan uses hardcoded cookies and works just fine.

Adamantcheese:
Weird. No idea then.

K1rakishou:
Alright. I added two new settings to enable/disable the new captcha window and to use the real or the default google cookie. In case of using the real cookie a GET request to the google.com will be made to retrieve the NID cookie (apparently it is the most important among all of them) then this cookie will be used for recaptcha requests until app restart. Also I added a button to manually refresh the cookie (sometimes it helps when you are stuck in the challenge hell). It does look like it helps a little bit when using the real google cookie.

Clover/app/src/main/res/values/strings.xml
```diff
@@ -577,4 +577,19 @@ Don't have a 4chan Pass?<br>
     "</string>
 
     <string name="settings_screen_theme">Themes</string>
+
+    <!-- Captcha settings -->
+    <string name="settings_captcha_group" tools:ignore="MissingTranslation">Captcha</string>
+    <string name="settings_use_new_captcha_window" tools:ignore="MissingTranslation">Use new captcha window for no-js captcha</string>
+    <string name="settings_use_real_google_cookies" tools:ignore="MissingTranslation">Use real google cookies instead of hardcoded ones</string>
+    <string name="settings_use_real_google_cookies_description" tools:ignore="MissingTranslation">When the real google cookies a GET request to the google.com will be executed to get the google cookies (NID) which will be used for captcha authentication. Why would you need them? Because the hardcoded ones sometimes will make you re-enter the captcha dozens of times. Those cookies will be updated automatically (when posting the first time after the app start). But it will also be possible to update them manually if necessary (for example, when the old ones give too many challenges in a row).</string>
```
K1rakishou: Fix message

K1rakishou:
Instead of refreshing google cookies (when this option is enabled) on every app start I decided to do it once in three months (but there is still an ability to refresh them manually). I think it is a better idea to not refresh them way too often.

PR #707, opened 2019-05-18
#434, Add ability to sort boards by code/name in ascending/descending order

### LINKED ISSUE #434: Feature request: Sorting mode for boards
Please add a button or option to sort boards by their name instead of just their code.

K1rakishou:
Closes #434

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/presenter/BoardSetupPresenter.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/presenter/BoardSetupPresenter.java
@@ -26,6 +26,8 @@ import org.floens.chan.ui.helper.BoardHelper;
 import org.floens.chan.utils.BackgroundUtils;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -56,6 +58,7 @@ public class BoardSetupPresenter implements Observer {
     private List<String> selectedSuggestions = new LinkedList<>();
 
     private String suggestionsQuery = null;
+    private final BoardSortingInfo boardSortingInfo = BoardSortingInfo.defaultSorting();
 
     @Inject
     public BoardSetupPresenter(BoardManager boardManager) {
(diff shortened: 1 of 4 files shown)


Clover/app/src/main/java/org/floens/chan/core/presenter/BoardSetupPresenter.java
```diff
@@ -315,4 +347,120 @@ public long getId() {
...
+
+        private int options;
+
+        public BoardSuggestionComparator(BoardSortingInfo boardSortingInfo) {
+            options = boardSortingInfo.getSortingMode().value | boardSortingInfo.getSortingOrder().value;
+        }
+
+        @Override
+        public int compare(BoardSuggestion o1, BoardSuggestion o2) {
+            if (o1 == null && o2 == null) {
```
K1rakishou: Probably should just throw an exception when either of parameters is null.

Adamantcheese:
I think what was requested (which is also #703) is to sort SAVED boards, not boards that are being added. This would completely override the existing fuzzy search suggestion method.

K1rakishou:
>There was a sort A-Z option in the board editor, we should bring it back.

Isn't board editor the window where you add boards?

Adamantcheese:
Yeah, but with your current implementation the fuzzy search is completely ignored with no way to turn it back on. And either way there's no real reason to have it being sorted in any order with a search bar, just find what you want that way.

K1rakishou:
@Adamantcheese 
>And either way there's no real reason to have it being sorted in any order with a search bar, just find what you want that way.

Yeah, I guess it makes sense. But then I don't get what boards need to be sorted. What window exactly is it.

Adamantcheese:
I don't know either. People aren't being terribly specific with what boards they want sorted. I assumed that it was saved boards because that's what you look at the most?

K1rakishou:
By saved boards you mean this spinner?

Adamantcheese:
Yeah

K1rakishou:
Well, I guess this task can wait until the next release. I asked to elaborate on what exactly needs to be sorted.

PR #779, opened 2020-08-03
expand exoPlayer capabilities

Lolzen:
This lays groundwork mainly for other boards than 4chan which support audiofile uploading. also adds swf handling for /f/, e.g.

### DIFF
--- a/Clover/app/src/main/java/org/floens/chan/core/model/PostImage.java
+++ b/Clover/app/src/main/java/org/floens/chan/core/model/PostImage.java
@@ -23,7 +23,7 @@ import okhttp3.HttpUrl;
 
 public class PostImage {
     public enum Type {
-        STATIC, GIF, MOVIE
+        STATIC, GIF, MOVIE, SWF
     }
 
     public final String originalName;
@@ -57,7 +57,14 @@ public class PostImage {
                 break;
             case "webm":
             case "mp4":
+            case "mp3":
+            case "m4a":
+            case "ogg":
+            case "flac":
                 type = Type.MOVIE;
+                break;
+            case "swf":
+                type = Type.SWF;
                 break;
             default:
                 type = Type.STATIC;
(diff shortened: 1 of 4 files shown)


Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
```diff
@@ -443,6 +446,10 @@ public void onEnd() {
         });
     }
 
+    private void setOther(String fileUrl) {
+        Toast.makeText(getContext(), R.string.file_not_viewable, Toast.LENGTH_LONG).show();
```
floens, line 450: Maybe we can show this as text in the view itself, instead of a toast. If that's too much trouble leave it for another time.

Clover/app/src/main/res/values/strings.xml
```diff
@@ -138,6 +138,8 @@ Re-enable this permission in the app settings if you permanently disabled it."</
     <string name="image_open_failed">Failed to open image</string>
     <string name="image_spoiler_filename">Spoiler image</string>
 
+    <string name="file_not_viewable">File not internally viewable; you can still download this file.</string>
```
floens:
```suggestion
    <string name="file_not_viewable">File not internally viewable; you can still download this file or open it in your browser.</string>
```

Clover/app/src/main/java/org/floens/chan/core/model/PostImage.java
```diff
@@ -23,7 +23,7 @@
 
 public class PostImage {
     public enum Type {
-        STATIC, GIF, MOVIE
+        STATIC, GIF, MOVIE, SWF
```
floens, line 26: Maybe we can call it OTHER too since I see no future in playing swf files on Android.

Clover/app/src/main/java/org/floens/chan/ui/view/MultiImageView.java
Lolzen: I'm not opposed to that, i fact i think this is better. I'll try to find an example i could use to realize that, otherwise you're free to go ahead and change it to your liking.

Clover/app/src/main/java/org/floens/chan/core/model/PostImage.java
Lolzen: see PR # 780 
i could however change everything up if you believe it's a cleaner aproach.

floens: If there are apps for that maybe we can integrate it with those apps? But sure let's keep SWF then.

PR #800, opened 2021-06-21
Add 8kun and Fix Arisuchan support.

ghost:


ghost:
Fixed whatever seems broken to me.

